Issue #9, opened 2018-08-02
GPL-3.0 intake license not compatible with home-assistant Apache-2.0


alandtse:
If you're planning to upstream, you'll need to change your LICENSE to Apache-2.0 so that you can submit your addon later to the main tree.  The issue is anyone who offers you a pull request will assume you're taking their code as GPL-3.0, which cannot be converted by you to Apache-2.0 later without explicitly asking for permission to convert.  Asking to change licenses later is a pain particularly since you don't have a CLA in place.  VLC Media player went through this and you can see the pain in that experience [here](http://www.jbkempf.com/blog/post/2012/How-to-properly-relicense-a-large-open-source-project).  Based on the PR history, you seem clean of 3rd party code for now, so you may want to nip it in the bud.

Good job btw. Hope this succeeds.

keatontaylor:
solved with #13

PR #11, opened 2018-08-02
Fixes for JSONDecodeError and other misc issues.

### LINKED ISSUE #10: JSONDecodeError: Expecting value: line 8 column 1 (char 7)
Apologies in advance, I haven't fully fixed this yet, but wanted to report the issue and give you my findings.

As documented in the community [thread](https://community.home-assistant.io/t/echo-devices-alexa-as-media-player-testers-needed/), the Alexa component will fail to load with a JSONDecodeError.

The cause of the error is in AlexaAPI.get_devices because a valid JSON is not being returned from the [Alexa device page](https://alexa.amazon.com/api/devices-v2/device), but instead an Amazon sign-on page is being returned.

This is because the login method has an incorrect test.
```
            if 'devices' in post_resp.text
```
This fails because the sign in page does contain 'devices' so will give an incorrect true condition.  The test should instead check for a json, which apparently in python requires a [try clause](https://www.peterbe.com/plog/jsondecodeerror-in-requests.get.json-python-2-and-3). 
(edit fixed a bug in the code)
```
            try:
                    from json.decoder import JSONDecodeError
            except ImportError:
                    JSONDecodeError = ValueError
            try:
                post_resp.json()['devices']
            except JSONDecodeError:
                status['login_failed'] = True
            
            if ('login_failed' in status and status['login_failed']):
                _LOGGER.debug("Log in failure.")
            else:
                status['login_failed'] = False
                status['login_successful'] = True
                _LOGGER.debug("Succesfully logged in.")
```
This should fix the login test, but it doesn't fix it completely as the AlexaAPI.get_devices call may still get another sign-in page.  (This is where I had to stop; not sure why the existing login session isn't enough).  I think the solution would be to have the get_devices page (and probably all API calls) automatically login if it detects it's getting the sign-in page.

I'll probably try to implement it and will submit a PR if I get to it before you.

alandtse:
Fixes #10 by explicitly checking for the device json.  Still may make sense to have every API call check that the log-in is still valid although the use of cookies may moot that.

This contains other fixes and improvements. 

1.  Lxml was not in the requirements.
2. Captcha had a typo in the configurator.
3. session.[mainArt][url] had a keyerror. Added a check for key existence before assignment.
4. Saving of session cookies.
5. More debug logs.
6. Configurator failure message and explanation on what to do with blank images.

EDIT: Added more features and updated the PR description.

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -40,7 +40,7 @@ SUPPORT_ALEXA = (SUPPORT_PAUSE | SUPPORT_PREVIOUS_TRACK |
 _CONFIGURING = {}
 _LOGGER = logging.getLogger(__name__)
 
-REQUIREMENTS = ['beautifulsoup4==4.6.0']
+REQUIREMENTS = ['beautifulsoup4==4.6.0', 'lxml==4.2.3']
 
 MIN_TIME_BETWEEN_SCANS = timedelta(seconds=15)
 MIN_TIME_BETWEEN_FORCED_SCANS = timedelta(seconds=1)
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -126,7 +126,7 @@ def setup_alexa(hass, config, add_devices_callback, login_obj):
         """Update the devices objects."""
 
         devices = AlexaAPI.get_devices(url, login_obj._session)
-        devices = devices.json()['devices']
+        #devices = devices.json()['devices']
         bluetooth = AlexaAPI.get_bluetooth(url, login_obj._session).json()
 
         new_alexa_clients = []
@@ -528,6 +528,7 @@ class AlexaLogin():
         '''add username and password to the data for post request'''
         self._data[u'email'] = self._email
         self._data[u'password'] = self._password
+        self._data[u'rememberMe'] = "true"
 
         if captcha is not None:
             self._data[u'guess'] = captcha
@@ -548,10 +549,21 @@ class AlexaLogin():
             post_resp = self._session.get('https://alexa.' + self._url +
                         '/api/devices-v2/device')
 
-            if 'devices' in post_resp.text:
-                status['login_successful'] = True
-            else:
+            try:
+                from json.decoder import JSONDecodeError
+            except ImportError:
+                JSONDecodeError = ValueError
+            try:
+                post_resp_json = post_resp.json()['devices']
+            except JSONDecodeError:
                 status['login_failed'] = True
+
+            if ('login_failed' in status and status['login_failed']):
+                _LOGGER.debug("Log in failure.")
+            else:
+                status['login_failed'] = False
+                status['login_successful'] = True
+                _LOGGER.debug("Succesfully logged in.")
 
         self.status = status
 
@@ -669,8 +681,8 @@ class AlexaAPI():
         try:
             response = session.get('https://alexa.' + url +
                                    '/api/devices-v2/device')
-            return response
-        except:
-            _LOGGER.error("An error occured accessing the API")
+            return response.json()['devices']
+        except Exception as e:
+            _LOGGER.error("An error occured accessing the API", e)
             return None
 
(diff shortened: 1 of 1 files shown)


alandtse:
As an aside, given #9 which I think you should still resolve, you have my permission to upstream my changes as Apache-2.0.

media_player/alexa.py
```diff
@@ -12,6 +12,9 @@
 import requests
 import voluptuous as vol
 from bs4 import BeautifulSoup
+import pickle
+from threading import Lock
```
keatontaylor: Eventually imports not native to home assistant need to be be added to the requirements and loaded during the setup_platform phase. Although this will all be solved with a custom pip module eventually.

media_player/alexa.py
```diff
@@ -250,7 +272,7 @@ def refresh(self, device):
                 self._media_title = self._session['infoText']['title']
                 self._media_artist = self._session['infoText']['subText1']
                 self._media_album_name = self._session['infoText']['subText2']
-                self._media_image_url = self._session['mainArt']['url']
+                self._media_image_url = self._session['mainArt']['url'] if 'url' in self._session['mainArt'] else None
```
keatontaylor: I'd prefer to keep maximum line length to 80 characters. (I think this is the requirement for official components for HA.)

media_player/alexa.py
```diff
@@ -496,8 +530,8 @@ def get_inputs(self, soup):
                 pass
         return data
 
-    def login(self, cookies=None, captcha=None):
-
+    def test_loggedin(self, cookies=None):
```
keatontaylor: This can eventually be converted to an auth decorator (probably when the AlexaAPI is moved to a pip module)

media_player/alexa.py
keatontaylor: This thread locking will not pass the requirements to submit to HA as an official component. We need to avoid locking/blocking the component thread at all costs.

keatontaylor:
Thank you for the PR! I finished classes Monday so will be back to working on this soon.

media_player/alexa.py
alandtse: @keatontaylor The lock was probably unnecessary but came from the other HA component I lifted the code from. That was a sensor so a lock was probably critical on the file access. Should be fine to remove it.

media_player/alexa.py
alandtse: It is. I was being lazy to try to figure out if it resolved the Decode bug. A lot of the Amazon API access needs to be moved out to pypi anyway before this can be submitted so I wasn't running it though the HA linter.

media_player/alexa.py
alandtse: My understanding is pickle is actually native to python3. Threading is gone.

alandtse: Removed.

media_player/alexa.py
alandtse: Learned something new today about auth decorators. It probably needs to be for all the API functions as I suggested in #10. Agree that probably makes more sense in the pip module.

alandtse:
@keatontaylor I think I addressed your immediate concerns.  Let me know.

media_player/alexa.py
```diff
@@ -471,15 +493,28 @@ def device_state_attributes(self):
...
         self._email = email
         self._password = password
         self._session = None
         self._data = None
         self.status = None
-
-        self.login()
+        self._cookiefile = hass.config.path("{}.pickle".format(ALEXA_DATA))
+        cookies = None
+        if exists(self._cookiefile):
```
keatontaylor: hass.config.path already returns None if the file doesn't exist. So no need for the os.path exists() function here. 

You can simply do 

```
self._cookiefile = hass.config.path("{}.pickle".format(ALEXA_DATA))
cookies = None
if self._cookiefile:
```

alandtse:
@keatontaylor: Did some additional edits.

1. Version bumped to 0.8.0
2. Exists removed.
3. Found a Python 3.5 bug for JSONDecodeError and fixed it. 🤞
4. Fixed all HA linter errors since you want to submit this eventually.

PR #12, opened 2018-08-04
 Adding failure notification to configurator

alandtse:
This should help with some of the common complaints about blank images and whether login failed.


alandtse:
Closing as I have a larger PR that adds cookie support and this. It'll also replace my other one.

PR #22, opened 2018-08-08
Adding 2FA

alandtse:
Existing cookies appear to be accepted even swapping on 2FA so testing will require deleting the cookie.


FuzzyMistborn:
Tried this, got the following:
image

alandtse:
@FuzzyMistborn and you don't get the captcha regardless of how often you click confirm?  Can you please enable debugging on the component through home-assistant?

You'll have to do it through modifying your configuration.yaml for logger by setting the component logging to `debug`.

FuzzyMistborn:
Got it to work!  Not obvious that i had to click "configure" again but...i did and it asked for the security token and boom!

PR #23, opened 2018-08-10
Adding existence checks for nested dictionaries.

### LINKED ISSUE #21: TypeError: 'NoneType' object is not subscriptable
I managed to get as far as successfully configuring the component but then it gets the below error which repeats approximately every 50 seconds.
```
018-08-08 18:27:45 DEBUG (SyncWorker_14) [custom_components.media_player.alexa] Using cookies to log in.
2018-08-08 18:27:47 DEBUG (SyncWorker_14) [custom_components.media_player.alexa] Logged in.
2018-08-08 18:27:47 DEBUG (SyncWorker_14) [custom_components.media_player.alexa] Log in successful with cookies
2018-08-08 18:27:47 DEBUG (SyncWorker_14) [custom_components.media_player.alexa] Setting up Alexa devices
2018-08-08 18:27:54 ERROR (MainThread) [homeassistant.core] Error doing job: Future exception was never retrieved
  File "/usr/local/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/media_player/alexa.py", line 192, in setup_alexa
    update_devices()
  File "/usr/local/lib/python3.6/site-packages/homeassistant/util/__init__.py", line 325, in wrapper
    result = method(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/homeassistant/util/__init__.py", line 325, in wrapper
    result = method(*args, **kwargs)
  File "/config/custom_components/media_player/alexa.py", line 164, in update_devices
    update_devices, url)
  File "/config/custom_components/media_player/alexa.py", line 228, in __init__
    self.refresh(device)
  File "/config/custom_components/media_player/alexa.py", line 267, in refresh
    self._media_is_muted = self._session['volume']['muted']
TypeError: 'NoneType' object is not subscriptable
```
If there is any further info you require let me know.

Regards,
Michal

alandtse:
Should fix #21.

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -288,16 +288,35 @@ class AlexaClient(MediaPlayerDevice):
             self._session = self._session['playerInfo']
             if self._session['state'] is not None:
                 self._media_player_state = self._session['state']
-                self._media_pos = self._session['progress']['mediaProgress']
-                self._media_is_muted = self._session['volume']['muted']
-                self._media_vol_level = self._session['volume']['volume'] / 100
-                self._media_title = self._session['infoText']['title']
-                self._media_artist = self._session['infoText']['subText1']
-                self._media_album_name = self._session['infoText']['subText2']
+                self._media_pos = (self._session['progress']['mediaProgress']
+                                   if 'mediaProgress' in
+                                   self._session['progress'] else None)
+                self._media_is_muted = (self._session['volume']['muted']
+                                        if 'muted' in self._session['volume']
+                                        else None)
+                self._media_vol_level = (self._session['volume']
+                                                      ['volume'] / 100
+                                         if 'volume' in self._session['volume']
+                                         else None)
+                self._media_title = (self._session['infoText']['title']
+                                     if 'title' in self._session['infoText']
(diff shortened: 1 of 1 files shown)


keatontaylor:
Hmm, I was under the assumption that if state was not None that all the following keys would be in the json string even if they were simply None themselves or an empty string. I'd rather avoid adding a bunch of conditionals if it isn't absolutely necessary, but I am also unsure what the preferred design paradigm is for validating that a key/value pair exists or doesn't exist in a json string.

alandtse:
I come from a programming paradigm that treats all input data as suspect so that's how I would tend to solve this problem.  In HA based on my experience, I don't think there's a strong paradigm unless it's part of the PEP8 Style Guide.  What would you prefer to do in this situation?

mihalski:
I am firmly for simplicity/elegance but it's of little value practically when there is no functionality. 
Having said that I have no idea what the actual problem here is and thus I defer to those gracious enough to actually contribute something, even if it only works for some :P

PR #27, opened 2018-08-21
move api to pip package and add references to data and objects required

keatontaylor:
Merge request to move the API code to a PyPI module. @alandtse can you code-review this for me and give a thumbs up if you approve? 

The PyPI module can be found here: https://gitlab.com/keatontaylor/alexapy

Closes #27


alandtse:
Everything seems to work.  One thing we'll need to do is move out the hass object from the AlexaLogin and instead pass in the output directory.  Right now we're sending in hass instance which should be generalized.

keatontaylor:
I'll work on getting the hass stuff outside the component.

keatontaylor:
We should re-evaluate this considering the changes to the login that need to move to the PyPI package.

Issue #26, opened 2018-08-20
Feature Request: Device filter


dreimer1986:
Currently I have your 0.9.0 script working, mainly for TTS push messages on NAS device errors. Problem I have now is that my amazon account is polluted with devices I stopped using ages ago like a 1st gen Fire TV Stick and "Mobile Devies" and a "This device". A way to stop enlisting these would be very nice.

alandtse:
We may need more than an online filter for cases where Alexa may duplicate Sonos media_players.

mihalski:
Is any of the info in the JSON you mentioned sensitive? I can paste it in place of this message.

alandtse:
@mihalski: Yah the JSON is sensitive, so **don't** post it.  I think the deviceFamily and deviceFamily may be enough.

mihalski:
If there's a way I can provide the JSON to you privately I'd be happy to.

boojew:
Hey guys - I would love to have a filter (or explicit include) option. Anything I provide for this?

PR #30, opened 2018-08-25
Refresh only online devices

alandtse:
~~This should address #26~~
Since all devices are autodiscovered, it didn't make sense to create a per device filter in the configuration.  

~~This also enables a new optional configuration option which is by default on to only show online devices.~~
`media_player:
  - platform: alexa
    email: !secret amazon_user
    password: !secret amazon_pass
    url: amazon.com
    onlineonly: True`

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -222,7 +222,7 @@ def setup_alexa(hass, config, add_devices_callback, login_obj):
                                          update_devices, url)
                 alexa_clients[device['serialNumber']] = new_client
                 new_alexa_clients.append(new_client)
-            else:
+            elif device['online']:
                 alexa_clients[device['serialNumber']].refresh(device)
 
         if new_alexa_clients:


keatontaylor:
I don't agree with this functionality. I think we should exclude devices that are not capable of being used with the API, such as the "This Device", "Echo Sim", "Fire TV ", etc but the offline state is valuable if you have a device like in a car that remains offline but comes online when the vehicle is running. As for allowing the user to pick which devices he/she wants to see, I think they need to either group the items they want in the front-end of HA or they need to simply set the entities to hidden via customization. There is no advantage to excluding devices at the API level as we always get the entire list regardless.

keatontaylor:
I agree with not calling status information for off-line devices, I don't think I agree with excluding them on first launch. This means the user will have to figure out why those devices are not properly showing and results in more ambiguity. I say we remove the configuration option from the pull request and change it to just not update offline devices but still add them. Then users can hide devices they want with the HA customize.yaml file.

alandtse:
What if the default value was not to filter.  Users could then decide if they wanted to turn it off or not by explicitly changing the option?

alandtse:
Went ahead and changed the functionality per our discussion so only refreshes are impacted. Configuration is not changed.

PR #32, opened 2018-08-28
Fix one more "NoneType object is not iterable" issue

kwdavidson:
Fix one more "NoneType object is not iterable" issue.


alandtse:
Thanks. I actually have a more robust fix that tries to relogin as this None error should be the result of the login failing.  While the patch does stop the error, I believe everything actually stops working at this point.

keatontaylor:
Closing this pull request as pull #34 handles this.

PR #34, opened 2018-08-30
Handle Amazon verification pages

### LINKED ISSUE #33: Handle verification request
For users without 2FA, Amazon will occasionally request a sms or email verification. This was the cause of at least one Captcha loop report.

alandtse:
Should resolve #33.

Also minor fixes:
- Add debug option - This adds a debug option for sensitive data. It will actually write debug html files to the config directory and output emails/passwords to the debug log. While I recognize HA already has logger debug levels, I wasn't comfortable shipping a default version that would log passwords into HA. Should be easy to rip out, but I'm forced to manually add it in when I debug so I'd prefer to keep it.
- Fix modal orphan bug - This happens when we done a configurator before it closes.
- Capture Amazon error message to log and configurator
- Detect later Amazon login failure and automatically attempt relogin. Should address PR #32.
- Refactor login status testing code
- Minor fixes like updating user-agent

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -3,7 +3,7 @@ Support to interface with Alexa Devices.
 
 For more details about this platform, please refer to the documentation at
 https://community.home-assistant.io/t/echo-devices-alexa-as-media-player-testers-needed/58639
-VERSION 0.9.2
+VERSION 0.9.3
 """
 import logging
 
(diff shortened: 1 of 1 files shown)


alandtse:
Sorry about the massive amount of old commits that were previously committed. Is there a git command I should be using to squash accepted commits in my logs?

PR #45, opened 2018-11-20
Implement include/exclude device lists

nuttytree:
Keaton,

I got tired of seeing a bunch of extra devices in my list of Alexa devices so I added the ability to specify an include or an exclude list of devices by name.  This is my first time ever doing a pull request (in my professional life I work with Visual Studio Online so my Git experience is limited) so let me know if I did it wrong.

Chris

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -54,12 +54,16 @@ ALEXA_TTS_SCHEMA = MEDIA_PLAYER_SCHEMA.extend({
 })
 
 CONF_DEBUG = 'debug'
+CONF_INCLUDE_DEVICES = 'include_devices'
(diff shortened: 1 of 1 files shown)


alandtse:
I think this looks good. 
@keatontaylor I know we've previously discussed this feature request to have filtering at this level. If you've been following the thread, people keep asking for this. My vote would be to allow it since it's a configuration setting and the default behavior would be not to filter absent some optional configs.

osuapoc:
Side note, you may want to remove your versioning info at the top, or iterate it with this PR.

alandtse:
@osuapoc thanks. I bumped it.

PR #57, opened 2019-01-08
Add volume adjustment while in standby

### LINKED ISSUE #43: Possibility to control volume while media_player in standby mode
Thank you for your work put in to this component.

Please add possibility to change Echo device volume while it is in `standby` state.

Currently volume can be changed when media_player in Home Assistant is in state `playing` and `paused` but in order to make alexa_tts service constantly speak text at certain volume.

alandtse:
This should resolve #43. 

This is done by forcing a TTS for "......." before setting the volume if nothing is detected playing.  When playing a Flash Briefing, there is no media reported, but it appears to successfully change the volume without interrupting the skill.   Tested for US English so it's possible the multiple period may break in another language.

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -220,7 +220,7 @@ def setup_alexa(hass, config, add_devices_callback, login_obj):
                 continue
             elif exclude and device['accountName'] in exclude:
                 continue
-                        
+
             for b_state in bluetooth['bluetoothStates']:
                 if device['serialNumber'] == b_state['deviceSerialNumber']:
                     device['bluetooth_state'] = b_state
@@ -506,9 +506,11 @@ class AlexaClient(MediaPlayerDevice):
 
     def set_volume_level(self, volume):
         """Set volume level, range 0..1."""
+        if not self.available:
+            return
         if not (self.state in [STATE_PLAYING, STATE_PAUSED]
                 and self.available):
-            return
+            self.alexa_api.send_tts("..............")
         self.alexa_api.set_volume(volume)
         self._media_vol_level = volume
 


alandtse:
Do you know where to look? While my proposed fix is entirely a hack, all attempts to interface with the API are a hack. This is essentially doing what everyone is doing via an automation/script.

I'm also not sure the Routine would allow you to do it if it's in standby. It may require doing a TTS anyway.

keatontaylor:
I originally extracted some of the different API calls used by the Alexa android app by decompiling it. This was additional calls not included in the original shell script. We can look to see if the shell script has added volume control that is called via a routine (which should work regardless of the state of media playback) or I can decompile the latest android app and start digging.

alandtse:
Ok, I'll defer to you since I haven't reversed APKs before.

niemyjski:
@keatontaylor do you have any good guides for reversing apks. I'd like to look into what my Ecobee switches are doing with Alexa.

niemyjski:
Also currently I have a flow in node red that sets the volume on all media devices to 10% at night and 30% during the morning. It would be great if I could set this and have it queued in the background. As well as possibly queueing mute state?

alandtse:
@keatontaylor Would you object to using the TTS hack as a temporary fix until you have time to reverse the APK?

As for your other thought on routine sequences from the script, I looked when I implemented routines in the API but I didn't see an effective way to do it. While I could find examples of how to call existing sequences, I didn't see a programmatic way to create them. The script doesn't have a solution either for volume.  

Of course, I'll need to refactor this for #73 before I merge it in.

alandtse:
Nevermind, I found the sequence being used in routines `Alexa.DeviceControls.Volume`. This is tested to work even when set_media didn't. I'll close this proposal and add it to the 1.0.0 release.

PR #61, opened 2019-01-14
Added last_called boolean attribute

brianhanifin:
Instead of exposing each device's serial numbers as well as the last device's serial number for later comparison, per Keaton's suggestion (https://github.com/keatontaylor/custom_components/issues/60#issuecomment-453875723) I switched to a single `last_called` boolean attribute.

I have been testing this for a bit this afternoon... and the `last_called` boolean switches to the most recently activated Alexa device! Whether sending a TTS command via the component, or verbally speaking to an Alexa device, `last_called` updates across all Alexa devices within a few seconds!

This should satisfy keatontaylor#60

P.S. I also manually copied over the code that Keaton changed while I was working on my branch. Thank you for letting me contribute. If you have any changes you'd like me to make, let me know.

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -220,7 +220,7 @@ def setup_alexa(hass, config, add_devices_callback, login_obj):
                 continue
             elif exclude and device['accountName'] in exclude:
                 continue
-                        
+
             for b_state in bluetooth['bluetoothStates']:
                 if device['serialNumber'] == b_state['deviceSerialNumber']:
                     device['bluetooth_state'] = b_state
@@ -274,11 +274,17 @@ class AlexaClient(MediaPlayerDevice):
         # Class info
         self.alexa_api = AlexaAPI(self, session, url)
 
+        # Supports get_last_device_serial()
+        self.alexa_api_session = session
+        self.alexa_api_url = url
+        
         self.update_devices = update_devices
         # Device info
         self._device = None
         self._device_name = None
         self._device_serial_number = None
+        self._last_device_serial_number = None
+        self._last_called = None
         self._device_type = None
         self._device_family = None
         self._device_owner_customer_id = None
@@ -329,6 +335,8 @@ class AlexaClient(MediaPlayerDevice):
         self._source = self._get_source()
         self._source_list = self._get_source_list()
         session = self.alexa_api.get_state()
+        self._last_device_serial_number = self._get_last_device_serial()
+        self._last_called = self._get_last_called()
 
         self._clear_media_details()
         # update the session if it exists; not doing relogin here
@@ -417,10 +425,25 @@ class AlexaClient(MediaPlayerDevice):
                 sources.append(devices['friendlyName'])
         return ['Local Speaker'] + sources
 
+    def _get_last_device_serial(self):
+        last_device_serial = self.alexa_api.get_last_device_serial(self.alexa_api_url, self.alexa_api_session)
+        #last_device_serial = ""
+        return last_device_serial
+    
+    def _get_last_called(self):
+        if self._device_serial_number == self._last_device_serial_number:
+            return True
+        return False
+
     @property
     def available(self):
         """Return the availability of the client."""
         return self._available
+
+    @property
+    def last_device_serial(self):
+        """Return the last activity of the client."""
+        return self._last_device_serial_number
 
     @property
     def unique_id(self):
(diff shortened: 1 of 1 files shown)


brianhanifin:
Thanks for the push @alandtse. I have created a proper local GitHub repository from Keaton's master and am working on a proper merge to resubmit later.

brianhanifin:
I agree to contribute my code under Apache-2.0 license. Also, I will look into learning about flake8 listing.

This is my first foray into Python. A few years back I contributed to the [Nightscout Project](https://github.com/nightscout/cgm-remote-monitor). It has been a while since I have contributed to a project with other developers, so it may take me a bit to get back into the flow of things.

I appreciate getting the chance to contribute to this project. I love the TTS aspect and hope that adding the ability to identify which Alexa device was last active will help us craft Home Assistant TTS Scripts that can be routed to the Echo the user requested the information from. For example, @CCOSTAN's config includes a way to store and replay the last TTS message. He pipes that info across his whole house amp + speaker system. I have Echo's spread all over the house, and whenever I have a TTS message like that, I just pipe it to my Kitchen Echo Show because its is most central in thy downstairs portion of the house.

I will do some testing of my repeat last TTS message script before submitting again to make sure this will be reliable enough to do what I want it for.

alandtse:
Thanks Brian.  Please note flake8 is just a pylinter.  Many editors have a plug-in to enable it.  Or if you use pip, you can install flake8 directly to run it in the command line. It will check that your code is syntactically the same style as code that Home Assistant requires once we try to submit this officially (i.e., < 80 chars a line, proper tabs, etc.) .

We look forward to having your contribution when it's ready.

brianhanifin:
I setup the flake8 linter on Visual Studio Code and it is primarily giving me `Unable to import...` errors. But I suspect those can be ignored since I am coding on my Mac with the actual code being on my server. I believe the following errors are from the preexisting code, so I probably shouldn't be the one to troubleshoot those.

### Too many arguments for format string [E1305]
[Line: 927, Column: 27](https://github.com/brianhanifin/alexa_components/blob/9f0a9fc20a147e9c4db1ed449d97298c343d19c6/media_player/alexa.py#L927)
`_LOGGER.debug("Verification method requested: {}".format(claims_message, options_message))`
### Instance of 'AlexaAPI' has no '_device_owner_customer_id' member [E1101]
[Line: 1013, Column: 40](https://github.com/brianhanifin/alexa_components/blob/9f0a9fc20a147e9c4db1ed449d97298c343d19c6/media_player/alexa.py#L1013)
`else self._device_owner_customer_id) +`
### Instance of 'AlexaAPI' has no '_device_owner_customer_id' member [E1101]
[Line: 1037, Column: 40](https://github.com/brianhanifin/alexa_components/blob/9f0a9fc20a147e9c4db1ed449d97298c343d19c6/media_player/alexa.py#L1037)
`else self._device_owner_customer_id) +`

alandtse:
Those are mine. I'll handle those. If you're rebased, I think we can merge your in with @keatontaylor's approval.

media_player/alexa.py
```diff
@@ -1085,3 +1110,27 @@ def get_devices(url, session):
             _LOGGER.error("An error occured accessing the API: {}".format(
                 message))
             return None
+
+    @staticmethod
```
keatontaylor, line 1114: This does not have to be a static method. It is a call within the instantiated object of the Alexa Device class. It should be converted to a non-static method and the calls within the refresh method changed to use the instantiated alexa_api class instead of a static class member.

media_player/alexa.py
```diff
@@ -417,11 +425,26 @@ def _get_source_list(self):
                 sources.append(devices['friendlyName'])
         return ['Local Speaker'] + sources
 
+    def _get_last_device_serial(self):
```
keatontaylor, line 428: No need for a helper method that gets a value, then simply returns that same value. Just set it on line 338

media_player/alexa.py
```diff
@@ -417,11 +425,26 @@ def _get_source_list(self):
...
+        last_device_serial = self.alexa_api.get_last_device_serial(self.alexa_api_url, self.alexa_api_session)
+        #last_device_serial = ""
+        return last_device_serial
+    
+    def _get_last_called(self):
```
keatontaylor, line 433: This one is fine since it is handling a conditional.

PR #65, opened 2019-01-23
adding last_called

brianhanifin:
Following [these instructions](https://stackoverflow.com/a/9646323) I reset my repository, reapplied [`git config core.filemode false`](https://stackoverflow.com/a/1257613), then pasted my changes back into .gitignore + alexa.py.

Hopefully this puts the repo in a better state for you to work with.

### DIFF
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -1017,7 +1017,12 @@ class AlexaAPI():
             message = template.format(type(ex).__name__, ex.args)
             _LOGGER.debug("An error occured accessing the API: {}".format(message))
             return None
-        return response.json()['activities'][0]['sourceDeviceIds'][0]['serialNumber']
+        
+        # Ignore discarded activity records
+        if response.json()['activities'][0]['activityStatus'](0) != 'DISCARDED_NON_DEVICE_DIRECTED_INTENT':
+            return response.json()['activities'][0]['sourceDeviceIds'][0]['serialNumber']
+        else:
+            return None
 
     def play_music(self, provider_id, search_phrase, customer_id=None):
         """Play Music based on search."""
--- a/media_player/alexa.py
+++ b/media_player/alexa.py
@@ -443,10 +443,9 @@ class AlexaClient(MediaPlayerDevice):
             for devices in self._bluetooth_state['pairedDeviceList']:
                 sources.append(devices['friendlyName'])
         return ['Local Speaker'] + sources
-    
+
     def _get_last_called(self):
-        last_device_serial = self.alexa_api.get_last_device_serial()
-        if self._device_serial_number == last_device_serial:
+        if self._device_serial_number == self.alexa_api.get_last_device_serial():
             return True
         return False
 
@@ -454,11 +453,6 @@ class AlexaClient(MediaPlayerDevice):
     def available(self):
         """Return the availability of the client."""
         return self._available
-
-    @property
-    def last_device_serial(self):
-        """Return the last activity of the client."""
-        return self._last_device_serial_number
 
     @property
     def unique_id(self):
@@ -1009,18 +1003,19 @@ class AlexaAPI():
             _LOGGER.error("An error occured accessing the API: {}".format(
                 message))
             return None
+
         try:
-            response.json()
+            last_activity = response.json()['activities'][0]
         except Exception as ex:
             template = ("An exception of type {0} occurred."
                         " Arguments:\n{1!r}")
             message = template.format(type(ex).__name__, ex.args)
             _LOGGER.debug("An error occured accessing the API: {}".format(message))
             return None
-        
+
         # Ignore discarded activity records
-        if response.json()['activities'][0]['activityStatus'](0) != 'DISCARDED_NON_DEVICE_DIRECTED_INTENT':
-            return response.json()['activities'][0]['sourceDeviceIds'][0]['serialNumber']
+        if last_activity['activityStatus'](0) != 'DISCARDED_NON_DEVICE_DIRECTED_INTENT':
+            return last_activity['sourceDeviceIds'][0]['serialNumber']
         else:
             return None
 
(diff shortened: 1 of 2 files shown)


brianhanifin:
On a related note I experimented with this in Node-RED for a few minutes this morning. It appears I was having trouble identifying the last Alexa device to be called, because every Echo that can hear my voice reports back to the Activity API.

In the screenshot below you see that both my Kitchen Echo Show and my Family Room Echo both reported back. Unfortunately, the order in which these report back to our component is not consistent. Sometimes the last reported device is the one we want, and other times it reports back first.

screen shot 2019-01-23 at 8 54 41 am

For now this feature is not going to do what we want it to. However, there is a lot of data reported back by the Activity API, such as the exact phrase recognized by the Echo (or in the case of a non-command it reports that as well). I am hopeful that I will be able to find something in that stack that will help me make this feature reliable.

I am going to leave it alone for now so @alandtse can work on splitting them off.

niemyjski:
@brianhanifin LOL, that's going to be a problem. I have an Alexa in every room and some rooms have 3 devices.  Is there any way to see what one took the command. Sometimes I have two respond with one saying cannot complete and the other actually responding. 

It would be great to have additional inputs to say last command or last phrase or last action. For example you can say Alexa `I'm leaving` to put her into listening mode. Would be great to capture this and trigger HA to do something as well.

brianhanifin:
> It would be great to have additional inputs to say last command or last phrase or last action. For example you can say Alexa `I'm leaving` to put her into listening mode. Would be great to capture this and trigger HA to do something as well.

@niemyjski I am hoping to use this feature myself. For example I'd like to say Goodbye to Alexa and have the Echo I spoke to respond with [a random line from my `goodbye_line()` macro](https://github.com/brianhanifin/Home-Assistant-Config/blob/master/templates/speech/briefing.yaml#L12).

Here is a link to a couple of quick tests of the Activity API. I think I can code it to ignore `"activityStatus":"DISCARDED_NON_DEVICE_DIRECTED_INTENT"`.
https://gist.github.com/brianhanifin/f5595578015582a4c22f81827b28af80

brianhanifin:
OK, so after doing a few experiments... I came to the conclusion that ignoring the `DISCARDED_NON_DEVICE_DIRECTED_INTENT` activity status devices seem to make it more reliable. It was so easy to add a quick if statement that I pushed it in. I'm going to play with some automations and see if I can get the right Alexa device to play the response. 😀 

https://github.com/keatontaylor/custom_components/pull/65/commits/c19f489a4cf5f9b1ebc98008970b8a49836bc1d1

niemyjski:
Awesome work! If you are in discord hit me up. I wonder if this last_called could also be used for room awareness too or if there is any events that are being submitted for audio activity (I usually keep it on mute) but would be good to trigger a binary_sensor for audio detected or audio playing :) If you are close and you say something it may give you a better idea of proximity even if it's `DISCARDED_NON_DEVICE_DIRECTED_INTENT`

alandtse:
Thanks. This looks good. I'll merge it in.

media_player/alexa.py
```diff
@@ -436,12 +443,23 @@ def _get_source_list(self):
...
 
     @property
     def available(self):
         """Return the availability of the client."""
         return self._available
 
+    @property
+    def last_device_serial(self):
+        """Return the last activity of the client."""
+        return self._last_device_serial_number
```
alandtse: Where is this value set?

media_player/alexa.py
```diff
@@ -436,12 +443,23 @@ def _get_source_list(self):
...
+        return False
 
     @property
     def available(self):
         """Return the availability of the client."""
         return self._available
 
+    @property
+    def last_device_serial(self):
+        """Return the last activity of the client."""
```
alandtse: Rephrase the comment to match what we're returning.

media_player/alexa.py
```diff
@@ -979,6 +998,32 @@ def _get_request(self, uri, data=None):
...
+            response = self._get_request('/api/activities?startTime=&size=1&offset=1')
+        except Exception as ex:
+            template = ("An exception of type {0} occurred."
+                        " Arguments:\n{1!r}")
+            message = template.format(type(ex).__name__, ex.args)
+            _LOGGER.error("An error occured accessing the API: {}".format(
+                message))
+            return None
+        try:
+            response.json()
```
alandtse: Within the try clause, let's save the json portion we need to a variable and use the variable later in lines 1022, 1023. This will help readability and also should catch a potential exception if the json array is empty.  For example something like:
```suggestion
            last_activity = response.json()['activities'][0]
```
Please review the suggested variable name; I'm not certain what's the best way to describe it.

media_player/alexa.py
brianhanifin: I believe this is was used in my first round. I thought I had removed all of that code. Thanks for catching it.

media_player/alexa.py
brianhanifin: see below

media_player/alexa.py
brianhanifin: Good idea. It should make it a bit easier to follow the json tree as well. Thank you.

alandtse: I mean let's move it into the first try block. No need to duplicate the exception catching code.

alandtse:
No need to apologize. You're helping in your free time.

media_player/alexa.py
```diff
@@ -979,6 +992,33 @@ def _get_request(self, uri, data=None):
...
+        except Exception as ex:
+            template = ("An exception of type {0} occurred."
+                        " Arguments:\n{1!r}")
+            message = template.format(type(ex).__name__, ex.args)
+            _LOGGER.error("An error occured accessing the API: {}".format(
+                message))
+            return None
+
+        try:
+            last_activity = response.json()['activities'][0]
```
alandtse, line 999: @brianhanifin Not sure if you saw my comment as you closed the convo, but let's combine this with the prior try block so we're not duplicating the catch.

brianhanifin:
I have added the following to the wiki page.

### Last called device

Each device will report whether it is the last_called or not. This could allow us to identify the device that was called according to the Alexa Activities API.

### Creating sensor.last_alexa

```
sensor:
  - platform: template
    sensors:
      last_alexa:
        value_template: >
          {%- for entity in states.media_player -%}
            {%- if state_attr(entity.entity_id, 'last_called') == True -%}
              {{ entity.entity_id }}
            {%- endif -%}
          {%- endfor -%}
```

niemyjski:
Do you mind sharing an export for your node red script above.

PR #73, opened 2019-02-08
Update to breaking feature branch 1.0.0

alandtse:
This requires 0.1.0 of [alexapy](https://gitlab.com/keatontaylor/alexapy/merge_requests/2) and **HA >= 0.88.0**.

[Changes](https://github.com/keatontaylor/custom_components/wiki/Changelog#100-in-dev).
* Split to platform and pypi API (resolves #14, #62)
* Add multiple account support (resolves #37)
* Fix bug where cookie login didn't detect changes in configuration (resolves #63)
* Reduce media player polling. Polling should restart using HA or voice, but not other Apps. (resolves #70)
* Add [sequence processing](https://github.com/keatontaylor/custom_components/wiki#sequence-commands-versions--100) for more programmatic support.
* Add [automation routine](https://github.com/keatontaylor/custom_components/wiki#automation-routines-versions--100) support.
* Allow for volume changes when media not playing or paused (resolves #43)
This tested in both my production and test systems. I have updated wiki too in preparation for this.

PR #81, opened 2019-02-20
Fixing folder names for 0.88

TheFuzz4:
Updated the folder naming structure for HA 0.88.

### DIFF
--- a/alexa/media_player.py
+++ b/alexa/media_player.py
@@ -1,0 +1,1161 @@
+"""
+Support to interface with Alexa Devices.
+
+For more details about this platform, please refer to the documentation at
+https://community.home-assistant.io/t/echo-devices-alexa-as-media-player-testers-needed/58639
+VERSION 0.10.2
+"""
+import logging
+
+from datetime import timedelta
+
+import requests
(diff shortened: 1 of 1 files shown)


RowdyDog12:
This would be a great idea!

alandtse:
Thanks for putting this together.  However, as we are about to release a major rearchitecture as 1.X.X, I think we may not incorporate it into the 0.X.X branch to maintain some backward compatibility with versions of HA < 0.88.0. 

We may done one more point maintenance release as I think the fixes for 0.88.0 may not be backward compatible, and if we do, we'll see if we can merge this in.

alandtse:
One other comment, we had to avoid the `alexa` base name as it will conflict with the existing component.  Per the architecture change announcement, you need to completely override the existing component so this fix will break at some point in the future since we are not ready to integrate into the `alexa` component yet.

In the 1.0.0 branch (#73), we are using `alexa_media` for now.

TheFuzz4:
Cool no worries :) . just wanted to contribute.  I'm also going to rename my local to alexa_media as that is good to know.  Thanks.

PR #136, opened 2019-03-25
Merge alandtse/dev

### LINKED ISSUE #124: Error task exception was never retrieved (entity ID already exist)
Version 1.2.1; got the following error entries. HA was already running for a number of hours with Alexa working.

```
2019-03-14 05:14:03 ERROR (SyncWorker_12) [alexapy.alexaapi] An error occured accessing AlexaAPI: An exception of type ConnectionError occurred. Arguments:
(ProtocolError('Connection aborted.', ConnectionResetError(54, 'Connection reset by peer')),)
2019-03-14 05:14:09 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
  File "/Users/Media/homeassistant/lib/python3.7/site-packages/homeassistant/helpers/entity_platform.py", line 344, in _async_add_entity
    raise HomeAssistantError(msg)
homeassistant.exceptions.HomeAssistantError: Entity id already exists: media_player.ethan_s_echo. Platform alexa_media does not generate unique IDs
2019-03-14 05:14:09 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
  File "/Users/Media/homeassistant/lib/python3.7/site-packages/homeassistant/helpers/entity_platform.py", line 344, in _async_add_entity
    raise HomeAssistantError(msg)
homeassistant.exceptions.HomeAssistantError: Entity id already exists: media_player.master_bedroom. Platform alexa_media does not generate unique IDs
```

It was reporting the above for every Echo I have. Looks like loss of connection, reconnects but then when reconnects it tries to create new entities. 
I also checked and it is all still working. Hence reconnecting itself went fine with existing entity IDs working.

### LINKED ISSUE #132: web socket causes MIN_TIME_BETWEEN_FORCED_SCANS to be enforced
Currently a player state change remotely will work as intended, the websocket will trigger an update/refresh methods and updates the player state on the HA GUI.

However, this is not true for users who hit the play/pause button within the HA GUI now, the state is instead left as play after hitting the play button or pause. This state will not update since if the web socket connection is established it will not poll.

Problem seems to be that the self.update() call is getting called both when the user hits play and again when the websocket message comes in, this rapid succession triggers the MIN_TIME_BETWEEN_FORCED_SCANS and prevents the player from updating accordingly. Ideally, we should only trigger refreshes that poll amazon's API from websocket events when a websocket is established for those things that can be updated from user behavior.

alandtse:
Per our discussion in #132, this completely disables polling if websocket is detected. It also will do new device detection based on websockets. Should resolve #132. 

It does **not** handle bluetooth changes so is not a complete fix until your bluetooth update changes are merged in.  In theory you could simply force update_devices(no_throttle=True) after setting the `new_devices` key I mentioned in the other thread and the bluetooth update will be done as well.  Alternatively, you could move the bluetooth update over to it's own function which you would then call.

Also added some miscellaneous changes:

- Fix bug where reconnects would add known devices again (fix #124)
- Fix use of DATA_ALEXAMEDIA instead of DOMAIN to comply with HA requirements data store requirements
- Fix bug where we'd spawn multiple update_devices updates on reconnect

### DIFF
--- a/alexa_media/__init__.py
+++ b/alexa_media/__init__.py
@@ -481,8 +481,8 @@ def setup_alexa(hass, config, login_obj):
          ['entities']) = {'media_player': {}}
         (hass.data[DATA_ALEXAMEDIA]
          ['accounts'][email]['new_devices']) = True  # force initial update
+        track_time_interval(hass, lambda now: update_devices(), scan_interval)
     update_devices()
-    track_time_interval(hass, lambda now: update_devices(), scan_interval)
     hass.services.register(DOMAIN, SERVICE_UPDATE_LAST_CALLED,
                            last_call_handler, schema=LAST_CALL_UPDATE_SCHEMA)
 
--- a/alexa_media/__init__.py
+++ b/alexa_media/__init__.py
@@ -416,8 +416,8 @@ def setup_alexa(hass, config, login_obj):
                                               hide_email(email)))[0:32],
                               {'player_state': json_payload})
             elif command == 'PUSH_DOPPLER_CONNECTION_CHANGE':
-                # Player volume update
-                _LOGGER.debug("Updating media_player avalibility %s",
+                # Player availability update
+                _LOGGER.debug("Updating media_player availability %s",
                               json_payload)
                 hass.bus.fire(('{}_{}'.format(DOMAIN,
                                               hide_email(email)))[0:32],
--- a/alexa_media/__init__.py
+++ b/alexa_media/__init__.py
@@ -402,44 +402,47 @@ def setup_alexa(hass, config, login_obj):
             _LOGGER.debug("%s: Received websocket command: %s : %s",
                           hide_email(email),
                           command, json_payload)
+            serial = None
             if command == 'PUSH_ACTIVITY':
                 #  Last_Alexa Updated
+                serial = (json_payload
+                          ['key']
+                          ['entryId']).split('#')[2]
                 last_called = {
-                    'serialNumber': (json_payload
-                                     ['key']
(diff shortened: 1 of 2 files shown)


alexa_media/__init__.py
```diff
@@ -257,14 +259,20 @@ def update_devices():
...
         devices = AlexaAPI.get_devices(login_obj)
         bluetooth = AlexaAPI.get_bluetooth(login_obj)
         _LOGGER.debug("%s: Found %s devices, %s bluetooth",
                       hide_email(email),
                       len(devices) if devices is not None else '',
                       len(bluetooth) if bluetooth is not None else '')
         if ((devices is None or bluetooth is None)
-                and not hass.data[DOMAIN]['accounts'][email]['config']):
+                and not hass.data[DATA_ALEXAMEDIA]
+                                 ['accounts'][email]['config']):
```
keatontaylor, line 275: I'd prefer, but perhaps in the next version to remove the scan interval stuff and throttle to make this entirely event driven. Should clean up the code considerably when we can just fire on events.

alexa_media/media_player.py
```diff
@@ -520,7 +518,9 @@ def set_volume_level(self, volume):
             return
         self.alexa_api.set_volume(volume)
         self._media_vol_level = volume
-        self.update()
+        if not (self.hass.data[DATA_ALEXAMEDIA]
```
keatontaylor, line 521: Same for this, assumption is websockets will always be active, let's make sure to remove these lines of code in the future and rely solely on event messages.

Issue #135, opened 2019-03-24
Add PUSH_BLUETOOTH_STATE_CHANGE from websocket messages


keatontaylor:
Adds ability to update the connected bluetooth device along with updating the list of bluetooth devices on a PUSH_BLUETOOTH_STATE_CHANGE websocket event.

Should be a straight forward fix, rebasing off @alandtse manual update removal code.

keatontaylor:
- PUSH_BLUETOOTH_STATE_CHANGE (anything bluetooth changes per device)
  - ENTERED_PAIRING_MODE (enters pairing mode, including connecting to an existing paired device)
  - DEVICE_CONNECTED (bluetooth device connected, does not include device ID)
  - DEVICE_DISCONNECTED (bluetooth device disconnected, does not include device ID)
  - FIND_DEVICES_SCAN_COMPLETED (scan for new bluetooth devices from the app complete)

PR #138, opened 2019-03-26
Bluetooth Websocket 

keatontaylor:
Added updating bluetooth devices from WebSocket messages. Looks like for every single PUSH_BLUETOOTH_STATE_CHANGE the https://alexa.amazon.com website pulls the bluetooth state information so if they're doing it I don't think we need to mess with narrowing down the call a bit more. If we hit rate limits or users have issues we can tidy it up.

### DIFF
--- a/alexa_media/__init__.py
+++ b/alexa_media/__init__.py
@@ -284,9 +284,33 @@ def setup_alexa(hass, config, login_obj):
         for device in devices:
             if include and device['accountName'] not in include:
                 included.append(device['accountName'])
+                if 'appDeviceList' in device:
+                    for app in device['appDeviceList']:
+                        (hass.data[DATA_ALEXAMEDIA]
+                         ['accounts']
+                         [email]
+                         ['excluded']
+                         [app['serialNumber']]) = device
+                (hass.data[DATA_ALEXAMEDIA]
+                 ['accounts']
+                 [email]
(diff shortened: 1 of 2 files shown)


alandtse:
Resolves #135

keatontaylor:
> Generally I think this is ok. I wonder if we should reuse the update_bluetooth_state in the polling but don't think it's a big deal.
> 
> I did notice some bracket/parentheses movement. While I know we didn't conclude on pylint vs mypy, I'm pretty OCD if I see a warning in my editor and will probably revert it back the next time I make edits (assuming it triggers the warning). ;)

I think vscode got a little trigger happy with reformatting some of the line endings, feel free to change them back on your next PR.

Can you test this code just so we don't have a testing sample size of 1?

alandtse:
I normally don't have bluetooth, but I don't see code to add new bluetooth_pairings. Is the bluetooth_state only for when a previously paired device connects?

keatontaylor:
bluetooth_state encompasses all PUSH_BLUETOOTH_STATE_CHANGE including adding, deleting, disconnect and connect

alandtse:
Before merging this, also merge in my fix on top #139. 

For some reason I'm not able to get the source list to dynamically update when I pair or unpair a device. I think we should investigate it.

PR #139, opened 2019-03-26
Add websocket checks to only trigger events for non-excluded serials

alandtse:


alandtse:
Also fixed the source list bug.

PR #146, opened 2019-03-29
Update dev to 1.2.4

### LINKED ISSUE #143: Custom Updater - Component URLs wrong
When looking at the WIKI entry to add the addin to custom updater you can see the line: https://raw.githubusercontent.com/keatontaylor/alexa_media_player/next/custom_components.json

Problem is, the next branch is not existent anymore. Maybe you might change this to dev branch, but in that case the json is still linking to the master branch and needs fixing.

### LINKED ISSUE #145: Routines are not found when many routines
For users with many routines, routines are not always found. The root cause is that our use fo the [automations url](https://alexa.amazon.com/api/behaviors/automations) does not return all routines.  This may be result of a paging variable we are missing.

There apparently is a [routines url](https://alexa.amazon.com/api/behaviors/routines) that contains all information for the reporting user.

Fix: Need to check both locations or determine why automations is not complete

alandtse:
- Bumps alexapy version (fixes #145)
- Adding dev version for custom_updater (fixes #143)
- Bumps version for next release

### DIFF
--- a/alexa_media/__init__.py
+++ b/alexa_media/__init__.py
@@ -26,9 +26,9 @@ from .const import (
 
 # from .config_flow import configured_instances
 
-REQUIREMENTS = ['alexapy==0.4.2']
+REQUIREMENTS = ['alexapy==0.4.3']
 
-__version__ = '1.2.3'
+__version__ = '1.2.4'
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


alandtse:
@keatontaylor I'll be out of touch for at least a week. I think we're ready to do a version increment. You should be able to merge this in to dev and then into master. I'll update the wiki changelog data you can use.

PR #170, opened 2019-04-26
Change source_list to return bluetooth profiles with A2DP-SOURCE

### LINKED ISSUE #166: Echo Buttons (and possibly other Alexa Gadgets) showing on Sources list
The Sources list appears to pick up all Bluetooth devices connected to the Alexa device, and not just speakers or devices (such as mobile phones) using the Echo as a Bluetooth speaker. 
Since Alexa Gadgets (such as the Echo Buttons and the Echo Wall Clock) connect over Bluetooth, these are also being included in the list. For me that means that I have several Echo Buttons showing in the list.

Is it possible to filter this list so it only has speakers are included?
If the type isn't returned it may be possible to do it by name. I know Echo Buttons all have names starting with EchoBtn. I don't know about the Echo Wall Clock as this hasn't yet launched in the UK.

alandtse:
Fixes #166

### DIFF
--- a/alexa_media/media_player.py
+++ b/alexa_media/media_player.py
@@ -360,7 +360,9 @@ class AlexaClient(MediaPlayerDevice):
         sources = []
         if self._bluetooth_state['pairedDeviceList'] is not None:
             for devices in self._bluetooth_state['pairedDeviceList']:
-                sources.append(devices['friendlyName'])
+                if (devices['profiles'] and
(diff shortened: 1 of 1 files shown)


ronmichel:
How would it be possible to change between different A2DP-SINKS? I would not need them to show up in the source_list but able to activate them in a service call. Thanks.

alandtse:
@ronmichel `media_player.select_source` will still work.

PR #199, opened 2019-06-01
Clear alexa voice history

macbury:
I have added new service called `alexa_media.clear_history` that removes first 50 entries from [alexa history page](https://alexa.amazon.com/spa/index.html#settings/dialogs).

You can combine this with simple automation:

```yaml
- alias: Clear Alexa History
  trigger:
    platform: time_pattern
    minutes: '/15'
  action:
    - service: alexa_media.clear_history
```

Waiting for approve on: https://gitlab.com/keatontaylor/alexapy/merge_requests/29


alandtse:
Thanks. Unfortunately [HA doesn't want us to do raw IO calls](https://developers.home-assistant.io/docs/en/development_checklist.html) within custom components and instead wants it done in the underlying API. Can you split the API calls into a PR to [alexapy ](https://gitlab.com/keatontaylor/alexapy) please?  Also, please target the PR to dev so we can stage.

macbury:
OK.

PR #321, opened 2019-08-31
fix(bluetooth): resolve bluetooth update issues

### LINKED ISSUE #316: Bluetooth device missing in source list
Yep. Issue is fixed. Thanks for the fix.

There is one more thing about the bluetooth...it does not show in the source list.

image

yet, when I catch the state change in node-red and verify the json object, I can see the bluetooth as source...yet not in the source list.
image
can we fix this to?

_Originally posted by @danielmihalache1983 in https://github.com/keatontaylor/alexa_media_player/issues/309#issuecomment-526648169_

alandtse:
Fixes #316

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -467,7 +467,7 @@ async def setup_alexa(hass, config, login_obj):
             for b_state in bluetooth['bluetoothStates']:
                 if device_serial == b_state['deviceSerialNumber']:
                     device['bluetooth_state'] = b_state
-                return device['bluetooth_state']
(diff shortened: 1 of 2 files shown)


alandtse:
There's an issue where we may see a rate exceeded message from the api (get_bluetooth). Right now we just don't update state. We should probably do a retry. However, the question is whether it should be done in alexapy or in the alexa_media component.

alandtse:
We should hold off on merging this until we have a replacement connect to bluetooth device feature.

alandtse:
This can be merged now. Didn't change the source select service so it should still work for pairing

PR #341, opened 2019-09-04
fix: clean discovery_info of unneeded values prior to passing

### LINKED ISSUE #336: WARNING (Recorder) Event is not JSON serializable: <Event platform_discovered[L]:
**Describe the bug**
Recorder provides a json warning because of passing of discovered info.

**To Reproduce**
1. Enable recorder for platform_loaded events
2. Update to 2.1.0
3. See error

**Expected behavior**
No warning

**Screenshots**

**System details**
 - Home-assistant (version): 0.98.2
 - Hassio (Yes/No): No
 - alexa_media  (version from `const.py` or HA startup): 2.1.0
 - alexapy (version from `pip show alexapy` or HA startup): 1.0.2

**Additional context**
This only appears at startup for all four components.
* media_player
* switch
* notify
* alarm_control_panel

alandtse:
fixes #336

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -426,12 +426,19 @@ async def setup_alexa(hass, config, login_obj):
                       exclude_filter)
 
         if new_alexa_clients:
+            cleaned_config = config.copy()
+            cleaned_config.pop(CONF_SCAN_INTERVAL, None)
+            # CONF_SCAN_INTERVAL causes a json error in the recorder because it
+            # is a timedelta object.
+            cleaned_config.pop(CONF_PASSWORD, None)
+            # CONF_PASSWORD contains sensitive info which is no longer needed
             for component in ALEXA_COMPONENTS:
                 hass.async_create_task(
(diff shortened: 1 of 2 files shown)


custom_components/alexa_media/__init__.py
```diff
@@ -426,12 +426,19 @@ def hide_serial(item):
                       exclude_filter)
 
         if new_alexa_clients:
+            cleaned_config = config.copy()
+            cleaned_config.pop(CONF_SCAN_INTERVAL, None)
+            # CONF_SCAN_INTERVAL causes a json error in the recorder because it
+            # is a timedelta object.
+            cleaned_config.pop(CONF_PASSWORD, None)
```
minglecm, line 433: Thank you for taking care of this.   Would it make sense to pop this out of `config` (or zero it out) as soon as it's known that it's not needed?

alandtse: It'd only partially help I think. The config is passed around HA and I haven't investigated whether components are truly sandboxed from the config.

Issue #446, opened 2019-10-22
Invalid Date shown for reminders, timers, and alarms on Safari and iOS devices


mikeage:
**Describe the bug**
Lovelace's `entities` card shows `Invalid Date` for my upcoming events, even though they appear to have a valid time.

**To Reproduce**
`Alexa, set a reminder for ...`
Look at an `entities` card for `sensor.foo_next_reminder` (also alarm and timer, but the log here is a reminder).

**Expected behavior**
Proper time

**Screenshots**

**System details**
 - Home-assistant (version): 0.100.2
 - Hassio (Yes/No): no
 - alexa_media  (version from `const.py` or HA startup): 2.3.3 + #436
 - alexapy (version from `pip show alexapy` or HA startup): 

**Logs**
See https://pastebin.com/bu1EahUj

here's the card's `sorted_active` field:
```
[...]

PR #447, opened 2019-10-25
fix(sensor): return all sensor states as isoformat strings

mikeage:
This resolves issue #446, which had Safari displaying "Invalid Date"
instead of a proper countdown. The strings were being rendered using
Python's default datetime renderer, which did not put a "T" in between
the day and the hour, as required by ISO-8601 (and Safari's Date()
parser, but not Chrome's)


alandtse:
Thanks! Can you please rebase it on #442 so I don't have to do a conflict and target dev?

mikeage:
No, I can't, because you did already :-)

alandtse:
Actually can you do a pull --rebase on `dev`? I merged in my other changes.

PR #740, opened 2020-05-23
Extend SwitchEntity instead of SwitchDevice

tmchow:
SwitchDevice is now deprecated


alandtse:
Thanks for putting this together, I didn't see this before I fixed it #742. We also need to make sure we don't break people on older versions of HA so it was slightly more involved.

Regardless, appreciate you taking the time to share your PRs. Please feel free to submit more in the future. :)

PR #804, opened 2020-06-24
Create es.json

Guille-Mas-Coll:
Translation to Spanish


alandtse:
Thank you. I'll merge this in for the next release. We're actually switching to Lokalise if you'd like to join our project there in case we add any more things to translate. Explanation on how to join is [here](https://github.com/custom-components/alexa_media_player/wiki/Translations).

PR #946, opened 2020-09-25
fix: update strings.json to remove title key

hwikene:
got this warrning "Warning: G] [TRANSLATIONS] config.title key has been moved out of config and into the root of strings.json. Starting Home Assistant 0.109 you only need to define this key in the root if the title needs to be different than the name of your integration in the manifest."


alandtse:
Great. I thought I had removed it, but must have missed it.

PR #947, opened 2020-09-26
Improve GH actions

KTibow:
This combines HACS and hassfest into one file, and also adds some flake8 lint.


alandtse:
Thanks. I don't think we're flake8 linted. Did you check to see that yet?

KTibow:
I did check that. It also installs WPS (wemake-python-styleguide), to give you some recommendations for ideal code.

alandtse:
Do you mind setting the extra flake tests as optional so we don't kill the build process until I've had a chance to review/fix?

KTibow:
What do you mean? It always passes.

alandtse:
I hadn't confirmed that the new action didn't fail you if it had warnings. Since it looks like it doesn't do that from your forks runs, I'll go ahead and merge it in.

PR #984, opened 2020-10-13
Update lokalize workflow

KTibow:
This improves the Lokalise workflow.


alandtse:
Thanks. The test workflow isn't actually used and was used for some ci testing. We actually already do the lokalise download in the semantice_release workflow as it packages the release.

KTibow:
Got it. If you won't be using it in the future you could delete it.

PR #1113, opened 2021-01-18
feat: add last_called notify service target

### LINKED ISSUE #1104: Is a new service: notify.alexa_media_last_alexa possible?
This would be nice rather than having to code a templated entity...
```
service: media_player.play_media
data:
  entity_id: '{{ states.sensor.last_alexa.state }}'
  media_content_id: Alexa.GoodMorning.Play
  media_content_type: sequence
```

alandtse:
This requires HA to allow dynamic targets.
closes #1104

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -33,6 +33,7 @@ from homeassistant.const import (
     CONF_PASSWORD,
     CONF_SCAN_INTERVAL,
     CONF_URL,
+    EVENT_HOMEASSISTANT_STARTED,
     EVENT_HOMEASSISTANT_STOP,
 )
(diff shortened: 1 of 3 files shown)


cibernox:
@alandtse In what version of HA are dynamic targets added? Does this require any user configuration?

alandtse:
It's automatic. It works without dynamic targets although it's less efficient. Once HA allows it, it'll switch over to the more efficient version.

alandtse:
As a note, until HA's release next month, there is a 4 second delay for the notify target to fully update.

cibernox:
Thanks for the heads up. 

Is that 4s any worse than the time taken by the current workaround of defining a custom sensor in home assistant?

PR #1137, opened 2021-01-24
Add device information to notification event

stuart-c:
Currently it is difficult to understand which device triggered a notification event. This adds the device name which triggered the event

### DIFF
--- a/custom_components/alexa_media/sensor.py
+++ b/custom_components/alexa_media/sensor.py
@@ -210,7 +210,7 @@ class AlexaMediaNotificationSensor(Entity):
         )
         self.hass.bus.async_fire(
             "alexa_media_notification_event",
-            event_data={"email": hide_email(self._account), "event": self._active[0]},
+            event_data={"email": hide_email(self._account), "device": {"name": self._name}, "event": self._active[0]},
         )
 
     def _fix_alarm_date_time(self, value):


custom_components/alexa_media/sensor.py
```diff
@@ -210,7 +210,7 @@ def _trigger_event(self, time_date) -> None:
         )
         self.hass.bus.async_fire(
             "alexa_media_notification_event",
-            event_data={"email": hide_email(self._account), "event": self._active[0]},
+            event_data={"email": hide_email(self._account), "device": {"name": self._name}, "event": self._active[0]},
```
alandtse, line 213: Thanks. Should this be the self.name property? Or is _name enough to differentiate the sensor that changed?

alandtse (commented):
Hi, I'd like to include but want to make sure we use the right name for exposure. Since it's your use case, is there a reason you don't wait the media player name that the sensor is attached to?

custom_components/alexa_media/sensor.py
stuart-c: Ideally it would be the entity ID of the sensor and/or media player. But basically my use case would be to be able to receive the event and then check some attributes on the correct _next_timer sensor within the automation. At the moment the event data doesn't contain anything which I can map to an entity.

alandtse: Thanks. I just added both name and entity_id.

PR #1218, opened 2021-03-16
fix: use ha converter for local time

### LINKED ISSUE #1217: Issue with DST time change and Alarms
**Describe the bug**
I use "next alarm" to start various automations when we wake up each morning; to that end I created a template sensor with this template:

`{{ as_timestamp(states('sensor.terry_s_2nd_echo_dot_next_alarm')) | timestamp_custom('%H:%M') }}`

This has worked well for many months, but the time changed this past weekend and now if I set an alarm on the templated Echo device (for example) at 6:00am, the time shown by the template is 7:00, 1 hour ahead of whatever time I asked Alexa to set.

I’ve checked the Echo Devices, the server time (I run on Debian Supervised), and the docker containers. All report the correct time, so I think it’s either my template, or the integration.

**To Reproduce**
Steps to reproduce the behavior: (Substitute appropriate times/intervals to reproduce locally. The times/intervals I used are shown below)
1.  10:00am local time, I watched an alexa device from the HA Integrations page while issusing the following verbal commands:

2. Click Path: (in HA) Configuration-->Integrations--> Alexa Media Player--> xxDevices-->{Choose Alexa Device}-->watch Entities Panel while uttering commands below
3. Say:
    “Alexa, set an alarm for two pm”
    “Alexa, set a reminder for test at two pm”
    “Alexa, set a timer for 4 hours”

**Expected behavior**
I should have had “In 4 hours” (2:00pm local time) for each item, however, as you can see from the image, the alarm added an hour. However, the Next Alarm time was ahead by 1 hour.

**Screenshots**
IMG_3779

**System details**

I run Home assistant Supervised:

Host OS: Debian 10 
HA version: 
Supervisor: supervisor-2021.03.4
HACS: 1.11.3
Alexa Media Player: v3.8.4

 - Home-assistant (version): core-2021.3.4
 - Hassio (Yes/No): Yes, version supervisor-2021.03.4
 - alexa_media  (version from `const.py` or HA startup): v3.8.4
 - alexapy (version from `pip show alexapy` or HA startup): 1.24.4
 - Amazon [2FA is enabled] Yes.

**Logs**
Nothing generated in logs

**Additional context**
I run a supervised install onto of Debian 10

alandtse:
closes #1217

### DIFF
--- a/custom_components/alexa_media/sensor.py
+++ b/custom_components/alexa_media/sensor.py
@@ -374,7 +374,7 @@ class AlexaMediaNotificationSensor(Entity):
 
     def _process_state(self, value):
         return (
-            value[self._sensor_property].replace(tzinfo=LOCAL_TIMEZONE).isoformat()
+            dt.as_local(value[self._sensor_property]).isoformat()
             if value
             else STATE_UNAVAILABLE
         )


roussell:
Loaded the dev branch and can confirm that the latest change works as intended. "Next Alarm" now reports the correct time. Thanks!

Issue #1202, opened 2021-03-08
Temperature sensor


Tonylolo:
Hi

Is it possible to have value of temperature sensor from amazon device? Dot and dot plus.

Instead of a weather station, for interior of house it's a good alternative!

Thank you for development.

alandtse:
Yes, dupe of #197 . Likely not to do it though since it would require polling the sensor.

GTunney:
I've just updated my alexa media player, I have 2 4th gen echos, the temperature sensor is showing for one but the other one isn't.

Any suggestions as to why it's only added the entity for one device?

Tonylolo:
> I've just updated my alexa media player, I have 2 4th gen echos, the temperature sensor is showing for one but the other one isn't.
> 
> 
> 
> Any suggestions as to why it's only added the entity for one device?

Hi

How do you have this in HA ?

PR #1244, opened 2021-04-14
Support lights and the temperature sensor of the Echo device

### LINKED ISSUE #1237: Support lights(and other devices?) integrated into Alexa
I have several zigbee lights and motion sensors already integrated with my Echo. From the Alexa app I can control them and view the state of all of them. I would like for Alexa Media Player to expose them to Home Assistant.

I have already figured out the appropriate network requests and I'm certain that at least basic control over lights is possible. I would be willing to take a shot at the development work on this if the project is interested in the idea.

This isn't a problem, just a dream.

I would like devices I have already integrated with Alexa to be added as entities to Home Assistant.

Integrating my lights directly into Home Assistant.

**Additional context**
I want this for a few reasons:
1. I don't want to buy a zigbee radio when my Echo already has one.
2. Alexa is completely solid. If my Home Assistant setup falls over, I lose advanced features but not basic control of my devices.
3. I can avoid setting up the smart home skill and still get voice control of my lights
4. I can run Home Assistant in the cloud

blm126:
Closes #1237 and #1202.

This moves all of the phoenix calls being done by the alarm platform into the data update coordinator so that sensor/light platform can share. This involved refactoring of the guard code.

Device support intentionally excludes all Skill-based devices Alexa has access to as there is likely another Home Assistant integration that can better address those items. The light support is quite limited supporting only brightness and power state. Temperature sensor support is complete, but is restricted specifically to the sensor built-in to an Echo.

With the exception of the new guard code, everything should be hidden by a new option. Enabling this feature requires turning that option on and reloading the integration.

I assume there are likely changes needed here. In particular, I wasn't sure on the process for translations. I saw the other languages, but I don't speak them. Should I run my text through Google translate? I'm also completely new to coding for home assistant and struggled with the appropriate way to hide this behind an option. Its definitely hidden, but the "reload the component" step feels like I'm missing some obvious way to do this better.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -44,7 +44,7 @@ from homeassistant.helpers.update_coordinator import DataUpdateCoordinator, Upda
 from homeassistant.util import dt, slugify
 import voluptuous as vol
 
-from .alexa_entity import get_entity_data, parse_alexa_entities
+from .alexa_entity import get_entity_data, parse_alexa_entities, AlexaEntityData
 from .config_flow import in_progess_instances
 from .const import (
     ALEXA_COMPONENTS,
@@ -271,7 +271,12 @@ async def async_setup_entry(hass, config_entry):
             "websocket": None,
             "auth_info": None,
             "second_account_index": 0,
-            "alexa_entities": (),
+            "alexa_entities": {
+                "lights": [],
+                "guards": [],
+                "temperature_sensors": []
+            },
+            "should_get_network": True,
             "options": {
                 CONF_QUEUE_DELAY: config_entry.options.get(
                     CONF_QUEUE_DELAY, DEFAULT_QUEUE_DELAY
@@ -317,10 +322,11 @@ async def async_setup_entry(hass, config_entry):
         return True
     return False
 
+
 async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
     """Set up a alexa api based on host parameter."""
 
-    async def async_update_data():
+    async def async_update_data() -> Optional[AlexaEntityData]:
         """Fetch data from API endpoint.
 
         This is the place to pre-process the data to lookup tables
@@ -328,6 +334,9 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
 
         This will ping Alexa API to identify all devices, bluetooth, and the last
         called device.
+
+        If any guards, temperature sensors, or lights are configured, their
+        current state will be acquired. This data is returned directly so that it is available on the coordinator.
 
         This will add new devices and services when discovered. By default this
         runs every SCAN_INTERVAL seconds unless another method calls it. if
@@ -353,7 +362,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         ].values()
         auth_info = hass.data[DATA_ALEXAMEDIA]["accounts"][email].get("auth_info")
         new_devices = hass.data[DATA_ALEXAMEDIA]["accounts"][email]["new_devices"]
-        should_get_network = not hass.data[DATA_ALEXAMEDIA]["accounts"][email]["alexa_entities"]
+        should_get_network = hass.data[DATA_ALEXAMEDIA]["accounts"][email]["should_get_network"]
 
         devices = {}
         bluetooth = {}
@@ -404,6 +413,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
                 if should_get_network:
                     _LOGGER.debug("Alexa entities have been loaded. Prepared for discovery.")
                     hass.data[DATA_ALEXAMEDIA]["accounts"][email]["alexa_entities"] = parse_alexa_entities(optional_task_results.pop())
+                    hass.data[DATA_ALEXAMEDIA]["accounts"][email]["should_get_network"] = False
 
                 if entities_to_monitor:
                     entity_state = optional_task_results.pop()
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -253,12 +253,11 @@ async def async_setup_entry(hass, config_entry):
             "coordinator": None,
             "config_entry": config_entry,
             "setup_alexa": setup_alexa,
-            "devices": {"media_player": {}, "switch": {}},
+            "devices": {"media_player": {}, "switch": {}, "guard": [], "light": [], "temperature": []},
             "entities": {
                 "media_player": {},
                 "switch": {},
                 "sensor": {},
-                "temperature": [],
                 "light": [],
                 "alarm_control_panel": {},
             },
@@ -271,11 +270,6 @@ async def async_setup_entry(hass, config_entry):
             "websocket": None,
             "auth_info": None,
             "second_account_index": 0,
-            "alexa_entities": {
-                "lights": [],
-                "guards": [],
-                "temperature_sensors": []
-            },
             "should_get_network": True,
             "options": {
                 CONF_QUEUE_DELAY: config_entry.options.get(
@@ -379,21 +373,22 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         if new_devices:
             tasks.append(AlexaAPI.get_authentication(login_obj))
 
-        entities_to_monitor = []
-        for temp in hass.data[DATA_ALEXAMEDIA]["accounts"][email]["entities"]["temperature"]:
-            if temp.enabled:
-                entities_to_monitor.append(temp.alexa_entity_id)
+        entities_to_monitor = set()
+        for sensor in hass.data[DATA_ALEXAMEDIA]["accounts"][email]["entities"]["sensor"].values():
+            temp = sensor.get("Temperature")
+            if temp and temp.enabled:
+                entities_to_monitor.add(temp.alexa_entity_id)
 
         for light in hass.data[DATA_ALEXAMEDIA]["accounts"][email]["entities"]["light"]:
             if light.enabled:
-                entities_to_monitor.append(light.alexa_entity_id)
+                entities_to_monitor.add(light.alexa_entity_id)
 
         for guard in hass.data[DATA_ALEXAMEDIA]["accounts"][email]["entities"]["alarm_control_panel"].values():
             if guard.enabled:
-                entities_to_monitor.append(guard.unique_id)
+                entities_to_monitor.add(guard.unique_id)
 
         if entities_to_monitor:
-            tasks.append(get_entity_data(login_obj, entities_to_monitor))
+            tasks.append(get_entity_data(login_obj, list(entities_to_monitor)))
 
         if should_get_network:
             tasks.append(AlexaAPI.get_network_details(login_obj))
@@ -412,7 +407,8 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
 
                 if should_get_network:
                     _LOGGER.debug("Alexa entities have been loaded. Prepared for discovery.")
-                    hass.data[DATA_ALEXAMEDIA]["accounts"][email]["alexa_entities"] = parse_alexa_entities(optional_task_results.pop())
+                    alexa_entities = parse_alexa_entities(optional_task_results.pop())
+                    hass.data[DATA_ALEXAMEDIA]["accounts"][email]["devices"].update(alexa_entities)
                     hass.data[DATA_ALEXAMEDIA]["accounts"][email]["should_get_network"] = False
 
                 if entities_to_monitor:
@@ -812,6 +808,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
                     "timestamp": json_payload["timestamp"],
                 }
                 try:
+                    await coord.async_request_refresh()
                     if serial and serial in existing_serials:
                         await update_last_called(login_obj, last_called)
                     async_dispatcher_send(
@@ -819,7 +816,6 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
                         f"{DOMAIN}_{hide_email(email)}"[0:32],
                         {"push_activity": json_payload},
                     )
-                    await coord.async_request_refresh()
                 except (AlexapyConnectionError):
                     # Catch case where activities doesn't report valid json
                     pass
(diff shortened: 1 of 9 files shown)


custom_components/alexa_media/__init__.py
```diff
@@ -255,6 +258,8 @@ async def login_success(event=None) -> None:
                 "media_player": {},
                 "switch": {},
                 "sensor": {},
+                "temperature": [],
```
alandtse: We have historically tried to store only HA recognized sensors. Can this be easily moved to sensor? Admittedly it may complicate things since we'll have to check the device_class to determine what sensor it's for.

custom_components/alexa_media/alexa_entity.py
```diff
@@ -0,0 +1,142 @@
...
+def get_friendliest_name(appliance: Dict[Text, Any]) -> Text:
+    """Find the best friendly name. Alexa seems to store manual renames in aliases. Prefer that one."""
+    aliases = appliance.get("aliases", [])
+    for alias in aliases:
+        friendly = alias.get("friendlyName")
+        if friendly:
+            return friendly
+    return appliance["friendlyName"]
+
+def parse_alexa_entities(network_details):
```
alandtse: Please add typing for the input/output.

custom_components/alexa_media/__init__.py
```diff
@@ -266,9 +271,13 @@ async def login_success(event=None) -> None:
             "websocket": None,
             "auth_info": None,
             "second_account_index": 0,
+            "alexa_entities": (),
```
alandtse: We should probably define what would be appropriate to include in this. Are we saying anything returned as part of this API call? Also, are you storing a tuple? Shouldn't it be a dict? Note my later comment to get typing added to the declarations.

custom_components/alexa_media/__init__.py
```diff
@@ -554,6 +587,7 @@ async def async_update_data():
                     },
                 },
             )
+        return entity_state
```
alandtse, line 596: If we're changing the function scope, we should update the typing and docstring to explain what this does now and particularly why `entity_state` should be returned as opposed to the other values that are being updated here. If it's just a sign we've successfully updated, then perhaps a bool is appropriate. Or a dict with all values. I could be wrong as I wrote this years ago so maybe I was inconsistent before.

custom_components/alexa_media/alarm_control_panel.py
```diff
@@ -125,147 +125,30 @@ async def async_unload_entry(hass, entry) -> bool:
...
-                "%s: Error refreshing alarm_control_panel %s: %s",
+        self._attrs: Dict[Text, Text] = {}
+        _LOGGER.debug(
+                "%s: Guard Discovered %s: %s %s",
                 self.account,
-                self.name,
-                json.dumps(state_json["errors"]) if state_json else None,
+                self._friendly_name,
+                self._appliance_id,
+                self._guard_entity_id,
```
alandtse: For amp, we actually try to obfuscate unique identifiers if we can. At alexapy we can leave it raw. There should be an obfuscate function that will work here. I believe we left the last 5 digits.

custom_components/alexa_media/alarm_control_panel.py
```diff
@@ -351,11 +227,10 @@ def supported_features(self) -> int:
         return SUPPORT_ALARM_ARM_AWAY
 
     @property
-    def available(self):
-        """Return the availability of the device."""
-        return self._available
+    def assumed_state(self) -> bool:
+        return self._login.session.closed
```
alandtse: Is this the right test? You may be right but the component can recycle the connection on reconnect.

custom_components/alexa_media/alexa_entity.py
```diff
@@ -0,0 +1,142 @@
...
+import json
+import logging
+from typing import Any, Dict, Text
+
+from alexapy import AlexaAPI
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def has_capability(appliance: Dict[Text, Any], interface_name: Text, property_name: Text) -> bool:
```
alandtse, line 20: Please generate documentation for a public function like this.

alandtse (changes requested):
Thanks. As for translations, you just need to do English. We use lokalize and once you modify the strings.json it will appear as something for our translators to work on. You will need to modify strings.json though.

custom_components/alexa_media/__init__.py
```diff
@@ -774,6 +809,7 @@ async def ws_handler(message_obj):
                         f"{DOMAIN}_{hide_email(email)}"[0:32],
                         {"push_activity": json_payload},
                     )
+                    await coord.async_request_refresh()
```
alandtse: Will there be a race condition with the dispatcher above? We used to rely on the receiving entity to do the refresh. If they're waiting for the coordinator, should the async_dispatcher be moved after this call? It's also possible we don't even need to dispatch anymore because the coordinator can provide that automatically..

alandtse:
Also, please plan to add a wiki entry explaining this new functionality plus the fact you have to enable it. I'll add the link into the release notes for this added feature.

custom_components/alexa_media/__init__.py
blm126: I took a shot at this, but as expected it got complicated quickly. The issue wasn't actually the device_class, but the fact that the sensor map is keyed based on the media player id and the temperature sensor doesn't really have a media player id.

If you think the rule about HA recognized sensors is important, I think the best way forward is to change the structure here to be "sensor": {"temperature": [], "media_player_attached": {}}. Should I move forward with that?

custom_components/alexa_media/__init__.py
blm126: Yes, it should have been a dict. My intention was that it should include the alexa smart home entities that can be successfully turned into HA entities. So, all the parsing/filtering/categorization of the entities happens before it is stored and it doesn't include the media players as they are very different.

custom_components/alexa_media/__init__.py
blm126: Docstring updated.  I based this off the home assistant examples of DataUpdateCoordinator.  By returning the value here, it is automatically available to everything extending CoordinatorEntity(guard, temp, lights) without those needing a handle to the global data object.

custom_components/alexa_media/__init__.py
blm126: The async_dispatcher call is still necessary as it is still received by the media player entities. I tested and investigated the data race concern and I believe there isn't a problem. In general, I think async_request_refresh is safe to call all the time. it is debounced and delayed automatically.

custom_components/alexa_media/alarm_control_panel.py
blm126: This was the test used before. However, now that you have pointed it out, the DataUpdateCoordinator does make it possible to write a more robust test that avoids needing to even think about login.

I believe that this is now a better check, but it is a slight change in behavior. Previous to my changes, assumed_state() was true only if there was a login failure on the last refresh. Now, assumed_state() will be true if the last refresh failed for any reason.

custom_components/alexa_media/__init__.py
alandtse: Can we just swap the order so the request refresh is sent before informing the media players? Dealing with the race condition later would be painful since it would happen randomly for random people.

custom_components/alexa_media/__init__.py
alandtse: Interesting. So the temperature sensor isn't obviously attached to the relevant Alexa device serial? 

I don't think we need the second level key. If the sensor has a unique id, it can be used as a key in the sensor dict mapping to the entity. While right now we're storing notifications in the sensor dict, those make sense cause they're linked to a media_player. 

So structurally:
```json
"sensor": { 
"SERIAL_NUMBER": { "Alarm" :{"HA_ENTITY_1"}, "Timer": {}, "Reminder":{}},
"OTHER_UNIQUE_IDENTIFIER": {
  "whatever makes sense for that identifier": "HA_ENTITY_2"
 }
}
```
Right now we don't enumerate across all keys in the entities/sensor dict so it shouldn't hurt to add unique keys to it.

custom_components/alexa_media/__init__.py
alandtse: If this is the raw json data from the API, we currently store that in the `devices` dict (probably should have been named as json but that was a historical decision based off of the API call). Would it be possible to add it there? It's ok that it's filtered down to valid items.

custom_components/alexa_media/__init__.py
alandtse: Ok makes sense. Thanks for that update then.

custom_components/alexa_media/alarm_control_panel.py
alandtse: Makes sense. I believe HA uses that now for all the assumed_state calls with an update coordinator.

alandtse:
Feel free to resolve conversations that you believe are addressed and let me know when you think it's ready to merging.

custom_components/alexa_media/__init__.py
blm126: Order swapped.

custom_components/alexa_media/__init__.py
blm126: I dove into the json a little more. It appears the device serial id was hiding in there for the echo temperature sensor. I used that. Sadly, I don't think there is actually any guarantee that it will link up to a media player so there is some small chance a new device serial may get introduced solely for a temperature sensor. I checked through all the code and verified that shouldn't break anything if it does happen.

blm126:
I have added some initial documentation to the Wiki [here](https://github.com/custom-components/alexa_media_player/wiki#discover-and-control-devices-connected-to-an-echo).

I believe everything is working and I have addressed all the issues you brought up and this is ready to merge from my side. Let me know if you see anything else you think should change. 

Also, thanks for the prompt code reviews and helping this move forward! This integration was lots of fun already and getting this feature in is going to let me do a bunch of new things with HA.

alandtse:
Thanks for putting this together. This was a long requested feature that I actually couldn't test myself. We'll leave this in dev for a week and then I'll do a new feature release. You may want to ask people on the forum to test it out.

alandtse:
FYI, opened an issue on dev because the properties key probably doesn't exist somewhere even though options has not been enabled
#1249 please take a look

alandtse:
Oh, I fixed 1249 pretty easily.  I'll put in a PR for that.

However, I got this warning:
```
2021-04-17 22:29:36 WARNING (MainThread) [homeassistant.components.light] Light is deprecated, modify AlexaLight to extend LightEntity
```

Please convert.

Also, with the options, you can actually programmatically force a reload if you have the entry_id. Consider doing that.
```python
     await self.hass.config_entries.async_reload(self.config_entry.entry_id)
```

This will expose emulated_hue lights as lights. We should warn about that or figure out if we can identify emulated_hue lights and ignore them. HA will allow you to look at other domains so we should check the code to see if emulated_hue is enabled.

Lastly, I got many warnings about unique ids. We'll want to eventually resolve that if we can figure out what is causing it. Also note the warning about no data. We should see if that can be ignored after the first time it happens.

```bash
2021-04-17 22:31:00 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new light.alexa_media entity: light.dining_room_lights
2021-04-17 22:31:00 DEBUG (MainThread) [custom_components.alexa_media.alexa_entity] Coordinator has no data for 6********************************04d
2021-04-17 22:31:00 DEBUG (MainThread) [custom_components.alexa_media.alexa_entity] Coordinator has no data for 6********************************04d
2021-04-17 22:31:00 ERROR (MainThread) [homeassistant.components.light] Platform alexa_media does not generate unique IDs. ID 2ee676ce-34f4-4367-a959-39d38a024c21 already exists - ignoring light.kitchen_lights
2021-04-17 22:31:00 ERROR (MainThread) [homeassistant.components.light] Platform alexa_media does not generate unique IDs. ID af23f902-1169-4b20-836e-22b59fa8646d already exists - ignoring light.guest_room_light
2021-04-17 22:31:00 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new light.alexa_media entity: light.master_closet_light
2021-04-17 22:31:00 DEBUG (MainThread) [custom_components.alexa_media.alexa_entity] Coordinator has no data for f********************************94e
2021-04-17 22:31:00 DEBUG (MainThread) [custom_components.alexa_media.alexa_entity] Coordinator has no data for f********************************94e
2021-04-17 22:31:00 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new light.alexa_media entity: light.playstation
2021-04-17 22:31:00 DEBUG (MainThread) [custom_components.alexa_media.alexa_entity] Coordinator has no data for b********************************c39
2021-04-17 22:31:00 DEBUG (MainThread) [custom_components.alexa_media.alexa_entity] Coordinator has no data for b********************************c39
2021-04-17 22:31:00 ERROR (MainThread) [homeassistant.components.light] Platform alexa_media does not generate unique IDs. ID 6380364a-7bb2-4283-91c8-c701568980d4 already exists - ignoring light.stairs_switch
2021-04-17 22:31:00 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new light.alexa_media entity: light.raspberry_pi
```

curt7000:
> Closes #1237 and #1202.
> 
> 
> 
> This moves all of the phoenix calls being done by the alarm platform into the data update coordinator so that sensor/light platform can share. This involved refactoring of the guard code.
> 
> 
> 
> Device support intentionally excludes all Skill-based devices Alexa has access to as there is likely another Home Assistant integration that can better address those items. The light support is quite limited supporting only brightness and power state. Temperature sensor support is complete, but is restricted specifically to the sensor built-in to an Echo.
> 
> 
> 
> With the exception of the new guard code, everything should be hidden by a new option. Enabling this feature requires turning that option on and reloading the integration.
> 
> 
> 
> I assume there are likely changes needed here. In particular, I wasn't sure on the process for translations. I saw the other languages, but I don't speak them. Should I run my text through Google translate? I'm also completely new to coding for home assistant and struggled with the appropriate way to hide this behind an option. Its definitely hidden, but the "reload the component" step feels like I'm missing some obvious way to do this better.

Is there a way to selectively support an Alexa Skill here?  I have one Alexa Skill with no HA integration (Frigidaire Dehumidifier) that would be amazing to turn off with an automation when our TV is on.

I've got a workaround using your integration where I am sending a voice command to Alexa, but actual status of the unit would be helpful in other ways.

blm126:
The limitation around Alexa Skills is hard coded in the current implementation. There isn't anything in the Alexa API stopping support for skill based devices. However, I don't have any that I could test. Especially for a dehumidifier, I have no idea how Alexa would report that. I don't imagine it would be much different than lights, but its probably different enough that it would need someone with access to the actual device to implement it.

PR #1253, opened 2021-04-18
Detect and ignore lights created by emulated_hue

### LINKED ISSUE #1251: Emulated hue lights discovered as lights
**Describe the bug**
With the inclusion of light support, [emulated_hue](https://www.home-assistant.io/integrations/emulated_hue/) lights are being detected and exposed. They should be filtered out since they already should exist in HA.

**To Reproduce**
1. Enable emulated hue for any existing switch in HA
2. Enable lights in options
3. Lights will be added to HA

**Expected behavior**
Emulated hue lights should be ignored

**Screenshots**

**System details**
 - Home-assistant (version): 
 - Hassio (Yes/No): (Please note you may have to restart hassio 2-3 times to load the latest version of alexapy after an update. This looks like a HA bug).
 - alexa_media  (version from `const.py` or HA startup): dev
 - alexapy (version from `pip show alexapy` or HA startup):
 - Amazon [2FA is enabled](https://github.com/custom-components/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). We will not debug login issues if unanswered: 

**Logs**
Please provide [logs](https://github.com/custom-components/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component). We'll be most likely asking for them anyway.

**Additional context**

blm126:
Fixes #1251. This should do a reasonable job of ignoring lights that originated at emulated_hue. Sadly, after much testing Alexa eventually had my real Hue bulb and the emulated hue bulb identical in every way except manufacturer name.

I'll document the restriction that older Hue bulbs may be ignored if you have emulated_hue enabled.

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -41,12 +41,11 @@ def has_capability(
     return False
 
 
-def is_emulated_hue(appliance: Dict[Text, Any]) -> bool:
-    """Determine if an appliance may have originated from an emulated hue hub.
+def is_hue_v1(appliance: Dict[Text, Any]) -> bool:
+    """Determine if an appliance is managed via the Philips Hue v1 Hub.
 
-    This check will catch all emulated_hue devices, but it may also catch too much.
-    Its possible some older Hue bulbs may actually use this manufacturer name even though modern bulbs report "Philips".
-    Therefore, only believe this check if emulated_hue is actually turned on.
+    This check catches old Philips Hue bulbs and hubs, but critically, it also catches things pretending to be older
+    Philips Hue bulbs and hubs. This includes things exposed by HA to Alexa using the emulated_hue integration.
     """
     return appliance.get("manufacturerName") == "Royal Philips Electronics"
 
@@ -112,7 +111,7 @@ class AlexaEntity(TypedDict):
     id: Text
     appliance_id: Text
     name: Text
-    is_possibly_emulated: bool
+    is_hue_v1: bool
 
 
 class AlexaLightEntity(AlexaEntity):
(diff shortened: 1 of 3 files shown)


custom_components/alexa_media/light.py
```diff
@@ -88,7 +95,7 @@ def alexa_brightness_to_ha(alexa):
     return alexa / 100 * 255
 
 
-class AlexaLight(CoordinatorEntity, Light):
+class AlexaLight(CoordinatorEntity, LightEntity):
     """A temperature sensor reported by an Echo. """
```
alandtse: Just noticed the doc is wrong. Please correct.

custom_components/alexa_media/alexa_entity.py
```diff
@@ -41,6 +41,16 @@ def has_capability(
...
 
 
+def is_emulated_hue(appliance: Dict[Text, Any]) -> bool:
+    """Determine if an appliance may have originated from an emulated hue hub.
+
+    This check will catch all emulated_hue devices, but it may also catch too much.
+    Its possible some older Hue bulbs may actually use this manufacturer name even though modern bulbs report "Philips".
+    Therefore, only believe this check if emulated_hue is actually turned on.
+    """
+    return appliance.get("manufacturerName") == "Royal Philips Electronics"
```
alandtse, line 50: Shouldn't the check for emulated_hue be also here? Otherwise this function seems incomplete since it really isn't answering the question.

alandtse (changes requested):

custom_components/alexa_media/alexa_entity.py
blm126: I left the core logic and checks the same but renamed the function, property, and rewrote the comments. Hopefully, that makes the code much more obvious. My thinking is that this code is mostly concerned with parsing and the policy decision on what to do with old Hue stuff should be made by the platform itself.

alandtse:
Great thanks.

PR #1254, opened 2021-04-18
Automatically reload the integration when extended entity discovery is enabled

blm126:
As the title says, simplify the experience by auto-reloading. I'll update wiki when this merges.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -1186,6 +1186,7 @@ async def update_listener(hass, config_entry):
     """Update when config_entry options update."""
     account = config_entry.data
     email = account.get(CONF_EMAIL)
+    reload_needed: bool = False
     for key, old_value in hass.data[DATA_ALEXAMEDIA]["accounts"][email][
         "options"
     ].items():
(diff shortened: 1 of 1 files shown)


custom_components/alexa_media/__init__.py
```diff
@@ -1199,6 +1199,9 @@ async def update_listener(hass, config_entry):
                 hass.data[DATA_ALEXAMEDIA]["accounts"][email]["options"][key],
             )
 
+            if key == CONF_EXTENDED_ENTITY_DISCOVERY and new_value:
```
alandtse: Shouldn't the reload happen regardless if it changes regardless of value? For example when disabling?

alandtse: Looking at the code more, new_value is already guaranteed to be non None and a changed value here. It can just be if key == ..

alandtse (changes requested):

custom_components/alexa_media/__init__.py
alandtse: Ok also fixed a big where the key forced the reload prematurely. Please test.

blm126: Tested and working well

alandtse:
I'll plan to take this to release this weekend. Will you have any bandwidth to address issues if they come up?

blm126:
Yes, I'll have some time.

alandtse:
Ok, we're live now. Please watch the issues in the next few days to see if anything pops up. I can't really debug this unfortunately.

PR #1266, opened 2021-04-30
PT-BR localization

antunes49:
brazilian portuguese localization


alandtse:
Thank you. Please note we handle translation through lokalise so you should consider joining our team there. See [instructions](https://github.com/custom-components/alexa_media_player/wiki/Translations).

PR #1270, opened 2021-05-01
Make lights controls more responsive and support colors

### LINKED ISSUE #1264: Support light colors
I'm annoyed that setting the color of my lights via custom commands causes my Echo to talk. Even more annoying is that I can't use scenes with this workaround without some truly terrible hacks with input_booleans and lots of automations. 

I want the native home assistant controls for lights to work with lights added va amp. This includes manual control on dashboards, groups, service calls, scenes, etc.

The workaround I added to the wiki with custom commands. 

**Additional context**
The biggest challenge here is that the Alexa API is terrible around color control. It reports light colors in HSB and color temperatures in kelvin, but you can't actually set those values. You have to set colors or temperatures using a pretty small list of predefined keywords. That is to say that Alexa supports exactly 5 possible values for color temperature and about ~100 different colors.

blm126:
Closes  #1264. Add color and color temperature support for lights, plus this makes the UI feel more responsive in general by skipping some refreshes when possible.

The biggest issue I have here is a question about what minimum version of HA does AMP target? HA currently requires color supported to be advertised as both a supported_feature(partially deprecated) and as a supported_color_mode(2021.4+). The color_mode support is new in 2021.4. The hacs.json shows "0.106.0", but the poetry.lock is bringing in 2021.4.

I'm fully supporting both ways of reporting colors, so I all I really need in 2021.4 is a handful of string constant imports. I could hard code those if 0.106 is the correct minimum version.

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -7,9 +7,10 @@ For more details about this platform, please refer to the documentation at
 https://community.home-assistant.io/t/echo-devices-alexa-as-media-player-testers-needed/58639
 """
 
+from datetime import datetime
 import json
 import logging
-from typing import Any, Dict, List, Optional, Text, TypedDict, Union
+from typing import Any, Dict, List, Optional, Text, Tuple, TypedDict, Union
 
 from alexapy import AlexaAPI, AlexaLogin, hide_serial
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator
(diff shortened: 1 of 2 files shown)


alandtse:
Poetry shows what we're developing against. HACS is a hard block on any versions below.

If the version would break an older installation entirely, then we need to bump HACS. If the feature just doesn't show up, then we don't change the HACS version. I have historically tried to make everything fail gracefully but if we want to require the last two major releases, I'm ok with that.

blm126:
I can gracefully support older versions with just a few lines of code. Its a bit ugly, but I don't think its worth changing the minimum version or the policy for this.

alandtse:
Works for me. Ready for the merge?

blm126:
Yep, good to go.

PR #1271, opened 2021-05-01
Add an event "alexa_media_alarm_dismissal_event"

### LINKED ISSUE #1014: Alarm Dismissed
**NOTE: For a feature to be created it has to be doable in the Amazon app because we are basically replicating the app for HA. If you can't find it in the app, we probably can't create it here.**

I would like to be able to trigger an automation when an alarm is dismissed. Now it is possible to do with a template to see when an alarm would go off, but if the alarm is snoozed it will not know. One of the routine triggers is alarm dismissed, I would like to have a button or event fired off when an alarm is actually shut off, not just snoozed.

blm126:
Closes #1014. This does 4 things:

1. Add an event "alexa_media_alarm_dismissal_event" any time an alarm is dismissed.
2. Add a status attribute to notification sensors
3. Add a dismissed attribute to notification sensors. This includes all notification types. It is updated to the current time any time that sensor has its notification dismissed.
4. Change the alarm sensor so that a snoozed alarm is classified as "active".

4 is a small change in behavior. It seemed to make sense to me. A snoozed alarm is still active.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -70,7 +70,12 @@ from .const import (
     SCAN_INTERVAL,
     STARTUP,
 )
-from .helpers import _catch_login_errors, _existing_serials, calculate_uuid
+from .helpers import (
+    _catch_login_errors,
+    _existing_serials,
+    alarm_just_dismissed,
+    calculate_uuid,
+)
 from .notify import async_unload_entry as notify_async_unload_entry
 from .services import AlexaMediaServices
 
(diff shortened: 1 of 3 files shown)


blm126:
When this merges, I know I need to update the wiki to document all of this.

As a side note, I am still working on a way to support multiple alarm sensors and to support on/off actions for alarms. I'm still experimenting to see if I can come up with a nice solution.

alandtse:
Thanks looks good.

PR #1353, opened 2021-08-06
Home Assistant uses packaging~=21.0 now

edestecd:
This was needed to make alexa media work in the newest home assistant


alandtse:
Code scanning fail isn't caused by this PR. That said, this change probably will break older versions of HA. Probably should do >20.3 since packaging doesn't seem to follow semver.

edestecd:
I'll update it

edestecd:
lol Yea I was thinking that as I made the change

PR #1377, opened 2021-10-08
Update switch.py

mwav3:
Update to new MDI entry for do not disturb switch per Home Assistant update 2021.10.0

### DIFF
--- a/custom_components/alexa_media/switch.py
+++ b/custom_components/alexa_media/switch.py
@@ -309,7 +309,7 @@ class DNDSwitch(AlexaMediaSwitch):
     @property
     def icon(self):
         """Return the icon of the switch."""
-        return super()._icon("mdi:do-not-disturb", "mdi:do-not-disturb-off")
+        return super()._icon("mdi:minus-circle", "mdi:minus-circle-off")
 
     def _handle_event(self, event):
         """Handle events."""


custom_components/alexa_media/switch.py
```diff
@@ -309,7 +309,7 @@ def __init__(self, client):
     @property
     def icon(self):
         """Return the icon of the switch."""
-        return super()._icon("mdi:do-not-disturb", "mdi:do-not-disturb-off")
+        return super()._icon("mdi:minus-circle", "mdi:minus-circle-off")
```
chemelli74, line 312: I would try to add a check on HA version and load one or the other based on that.

Simone

alandtse: While we do have version checking in other parts, that really complicates the code and I would accept a fix without it since it's a cosmetic issue. We won't release this fix until we get closer to the cutoff date anyway to give people time to migrate.

However, the warning only talks about the off state. Does the on state icon actually exist?

chemelli74: Nope, both are renamed:

```
2021-10-08 19:25:48 WARNING (MainThread) [frontend.js.latest.202110060] Icon mdi:do-not-disturb-off was renamed to mdi:minus-circle-off, please change your config, it will be removed in version 2021.12.
2021-10-08 19:26:54 WARNING (MainThread) [frontend.js.latest.202110060] Icon mdi:do-not-disturb was renamed to mdi:minus-circle, please change your config, it will be removed in version 2021.12.
```

Simone

alandtse:
You know what, given the only downside is broken icons, I'm ok with merging this now. People with problems can do the version fix or can just update.

PR #1383, opened 2021-10-13
2021-10-12

alandtse:


custom_components/alexa_media/manifest.json
```diff
@@ -7,6 +7,6 @@
   "issue_tracker": "https://github.com/custom-components/alexa_media_player/issues",
   "dependencies": ["persistent_notification", "http"],
   "codeowners": ["@keatontaylor", "@alandtse"],
-  "requirements": ["alexapy==1.25.1", "packaging~=20.3", "wrapt~=1.12.1"],
```
edestecd, line 10: Why was this change reverted? It is now broken again on the latest Home Assistant
packaging>=20.3

PR #1551, opened 2022-03-28
Add dev container management

chemelli74:
Add dev container files to allow easier development for those that are used to VScode self contained environment.


alandtse:
Thanks!

PR #1554, opened 2022-04-01
Use EntityCategory enum instead of strings

### LINKED ISSUE #1553: Error with latest Beta core 2022.4.0b1
**Describe the bug**
image

Feedback from Discord suggests the following "That looks like custom components using the string instead of the enum for EntityCategory"

Worked without issue using - Home Assistant 2022.4.0b0

https://github.com/home-assistant/core/releases

DND switch is now not available:
image

**To Reproduce**

**Expected behavior**

**Screenshots**

**System details**
 - Home-assistant (version): Home Assistant 2022.4.0b1
 - Hassio (Yes/No): (Please note you may have to restart hassio 2-3 times to load the latest version of alexapy after an update. This looks like a HA bug).
 - alexa_media  (version from `const.py` or HA startup):  v3.11.0
 - alexapy (version from `pip show alexapy` or HA startup):
 - Amazon [2FA is enabled](https://github.com/custom-components/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). We will not debug login issues if unanswered: y

**Logs**
2022-04-01 10:33:08 ERROR (MainThread) [homeassistant.components.switch] Error adding entities for domain switch with platform alexa_media
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 382, in async_add_entities
    await asyncio.gather(*tasks)
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 510, in _async_add_entity
    entry = entity_registry.async_get_or_create(
  File "/usr/src/homeassistant/homeassistant/helpers/entity_registry.py", line 345, in async_get_or_create
    return self.async_update_entity(
  File "/usr/src/homeassistant/homeassistant/helpers/entity_registry.py", line 529, in async_update_entity
    raise ValueError("entity_category must be a valid EntityCategory instance")
ValueError: entity_category must be a valid EntityCategory instance

**Additional context**

mdegat01:
Specifying the entity category as a string doesn't work anymore starting in `2022.4`, you must use the `EntityCategory` enum instead. This fixes #1553 

The enums have been accepted since `2021.12` ([PR](https://github.com/home-assistant/core/pull/60720)) which covers your supported versions so it is safe to make immediately, don't have to wait for `2022.4`.

### DIFF
--- a/custom_components/alexa_media/switch.py
+++ b/custom_components/alexa_media/switch.py
@@ -11,6 +11,7 @@ from typing import List, Text  # noqa pylint: disable=unused-import
 
 from homeassistant.exceptions import ConfigEntryNotReady, NoEntitySpecifiedError
 from homeassistant.helpers.dispatcher import async_dispatcher_connect
+from homeassistant.helpers.entity import EntityCategory
 
 from . import (
     CONF_EMAIL,
@@ -314,7 +315,7 @@ class DNDSwitch(AlexaMediaSwitch):
     @property
     def entity_category(self):
         """Return the entity category of the switch."""
-        return "config"
+        return EntityCategory.CONFIG 
 
     def _handle_event(self, event):
         """Handle events."""
@@ -355,7 +356,7 @@ class ShuffleSwitch(AlexaMediaSwitch):
     @property
     def entity_category(self):
         """Return the entity category of the switch."""
-        return "config"
+        return EntityCategory.CONFIG 
 
 class RepeatSwitch(AlexaMediaSwitch):
     """Representation of a Alexa Media Repeat switch."""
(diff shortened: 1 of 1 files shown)


mdegat01:
Ignore the mentions. I copied the wrong PR link into other similar PRs 🤦

alandtse:
Thanks. I'm actually not seeing the HA requirement in our manifest.json. However, it should be pretty easy to support older versions.

Can you please wrap the import in a try and catch the ImportError? When this happens, you can probably just create a dummy `EntityCategory.CONFIG` with the value `config`.

mdegat01:
@alandtse it's not the manifest it's a HACS thing. You have the supported version listed here:
https://github.com/custom-components/alexa_media_player/blob/2b14c3bf65c93cd39cf9992c912b29dcc8f28fb2/hacs.json#L13
Hacs won't allow users to install this component if ha is not at least that version.

I can still wrap in a try if you're worried about non hacs installs though.

alandtse:
Oh you're right. Thanks for the reminder. Ok, no need to wrap it.

PR #1555, opened 2022-04-03
Add requirements management

chemelli74:


alandtse:
We use poetry for dependency management. Please adopt that. If you need to convert for the dev container, do that.

Issue #1512, opened 2022-02-10
Reauthentication error


larena1:
**Describe the bug**
Reauthentication is impossible. Opening the Amazon website from the addon will show an error.
It seems to be related to the cookiejar conversion between aiohttp and httpx.
Only way to get it going again is removing the integration and setting it up again.

**To Reproduce**
1. Reauthenticate

**Expected behavior**

**Screenshots**

**System details**
 - Home-assistant (version): latest
 - Hassio (Yes/No): (Please note you may have to restart hassio 2-3 times to load the latest version of alexapy after an update. This looks like a HA bug).
 - alexa_media  (version from `const.py` or HA startup): 3.10.15
 - alexapy (version from `pip show alexapy` or HA startup): 1.25.3 (authcaptureproxy 1.1.1)
 - Amazon [2FA is enabled](https://github.com/custom-components/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). We will not debug login issues if unanswered: yes

**Logs**
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/aiohttp/web_protocol.py", line 435, in _handle_request
    resp = await request_handler(request)
  File "/usr/local/lib/python3.9/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/usr/local/lib/python3.9/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/security_filter.py", line 60, in security_filter_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/forwarded.py", line 220, in forwarded_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/request_context.py", line 28, in request_context_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/ban.py", line 79, in ban_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/auth.py", line 219, in auth_middleware
    return await handler(request)
  File "/usr/src/homeassistant/homeassistant/components/http/view.py", line 137, in handle
    result = await result
  File "/config/custom_components/alexa_media/config_flow.py", line 1122, in wrapped
    return await cls.handler(request, **kwargs)
[...]

PR #1564, opened 2022-04-11
Fix cookie issues

quthla:
https://github.com/custom-components/alexa_media_player/issues/1512

I ran into this issue and this fixed it for me.

### DIFF
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -378,16 +378,7 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
         proxy_url = self.proxy.access_url().with_query(
             {"config_flow_id": self.flow_id, "callback_url": str(callback_url)}
         )
-        if self.login.lastreq:
-            self.proxy.last_resp = self.login.lastreq
-            self.proxy.session.cookies = self.login._session.cookie_jar.filter_cookies(
-                self.proxy._host_url.with_path("/")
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. Please also disable the UI option to switch to `Use Login Proxy method (2FA not required)` with an relogin attempt. Otherwise, that will be silently broken with this change.

alandtse:
Oh let me explain. So yes, the login proxy method is preferred. However, the error in #1512 is caused when the login attempts to switch from the old legacy python based login to the proxy method. By removing the check you have done, you have disabled that conversion path, but the UI will end up in a loop because it won't be able to process that path despite the UI selection. 

You can recreate that effect by using the `force_logout` service, and then restarting HA. It will automatically try to relogin and if you then try to switch to login proxy, it should give you that error.

quthla:
Which situations is the path meant for exactly? I'm aware I removed the lastreq stuff but as proxy method still worked, I figured the code might be superfluous after all.

alandtse:
You could switch to proxy after any data entry screen to resume the login attempt. We need to remove any such switches where state was previously preserved. Or perhaps we don't make that guarantee and always restart.

quthla:
I'd say let's leave it as it is now. The path was broken anyway which led to never being able to reauthenticate via proxy and that's fixed now.

alandtse:
Ok you've convinced me. Thanks!

PR #1582, opened 2022-04-28
Updated italian translation

Andrea9530:
Improved and added the remaining translations


alandtse:
Thanks for contributing, but can you make the changes in [Lokalise](https://github.com/custom-components/alexa_media_player/wiki/Translations) please? We use that to handle translations and any changes here will be overwritten.

alandtse:
Closing since this should be done in Lokalise which has a UI. Once done there, it automatically updates here on the next release.

PR #1639, opened 2022-06-12
Fix notify.alexa_media_last_called race condition

lxpollitt:
Fix for https://github.com/custom-components/alexa_media_player/issues/1638.

Note that with the current system event notification processing order, the `elif` branch in the new code was never taken during my testing. But it seemed more robust to include this in case there is a scenario where it is taken (that I did not hit during my testing) or if the event processing order ever changes.

This is my first time writing any Python code in very many years, so apologies in advance if there's a better way of coding this.


alandtse:
Thanks.

PR #1682, opened 2022-08-05
Fix Temperature always showing Celsius

kdvlr:
Updated Code to add native_unit_of_measurement and device_class. Now the displayed temperature will be as per global  preferences set by user in Settings->System->General


alandtse (approved):
Thanks!

PR #1739, opened 2022-09-28
Set default data `type` for notify to `tts`

### LINKED ISSUE #1738: Set `data: type: tts` as default for notify service
**NOTE: For a feature to be created it has to be doable in the Amazon app because we are basically replicating the app for HA. If you can't find it in the app, we probably can't create it here.**

First of all, awesome integration. Thank you for sharing.

I started using the app and wanted to use the notify service.
It raised an error message, that `data` is missing from the service call.

After a short google search, I realized, that I have to enter `type: tts` as data.

I think, that it would make things easier, if we set `type: tts` as default, in case no data value `type` is given.

We could also set `announce` as default type.

provinzio:
As proposed by #1738, this PR sets the default data value of the notify service `type` to `tts`.

Closes #1738

### DIFF
--- a/custom_components/alexa_media/notify.py
+++ b/custom_components/alexa_media/notify.py
@@ -318,7 +318,10 @@ class AlexaNotificationService(BaseNotificationService):
                             )
                         )
                 else:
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks!

PR #1754, opened 2022-10-18
Add ZigBee contact sensors support

cosimomeli:
Hi,
implemented support for ZigBee Door/Window opening sensors.
This sensor is connected as binary_sensor entity to Home Assistant.

An example of Appliance Details of my sensor:
[contact_sensor.txt](https://github.com/custom-components/alexa_media_player/files/9812176/contact_sensor.txt)

`detectionState` is the capability holding the opened/closed state with text values DETECTED and NOT_DETECTED (for opened and closed states respectively)

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -94,7 +94,7 @@ def is_light(appliance: Dict[Text, Any]) -> bool:
     """Is the given appliance a light controlled locally by an Echo."""
     return (
         is_local(appliance)
-        and "LIGHT" in appliance["applianceTypes"]
+        and "LIGHT" in appliance.get("applianceTypes", [])
         and has_capability(appliance, "Alexa.PowerController", "powerState")
     )
 
@@ -102,7 +102,7 @@ def is_contact_sensor(appliance: Dict[Text, Any]) -> bool:
     """Is the given appliance a contact sensor controlled locally by an Echo."""
     return (
         is_local(appliance)
-        and "CONTACT_SENSOR" in appliance["applianceTypes"]
+        and "CONTACT_SENSOR" in appliance.get("applianceTypes", [])
         and has_capability(appliance, "Alexa.ContactSensor", "detectionState")
     )
 
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -98,6 +98,13 @@ def is_light(appliance: Dict[Text, Any]) -> bool:
         and has_capability(appliance, "Alexa.PowerController", "powerState")
     )
 
+def is_contact_sensor(appliance: Dict[Text, Any]) -> bool:
+    """Is the given appliance a contact sensor controlled locally by an Echo."""
+    return (
+        is_local(appliance)
+        and "CONTACT_SENSOR" in appliance["applianceTypes"]
+        and has_capability(appliance, "Alexa.ContactSensor", "detectionState")
+    )
 
 def get_friendliest_name(appliance: Dict[Text, Any]) -> Text:
     """Find the best friendly name. Alexa seems to store manual renames in aliases. Prefer that one."""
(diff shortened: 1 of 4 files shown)


alandtse:
Wow thanks. I'll review and see if we can merge it in.

custom_components/alexa_media/alexa_entity.py
```diff
@@ -98,6 +98,13 @@ def is_light(appliance: Dict[Text, Any]) -> bool:
         and has_capability(appliance, "Alexa.PowerController", "powerState")
     )
 
+def is_contact_sensor(appliance: Dict[Text, Any]) -> bool:
+    """Is the given appliance a contact sensor controlled locally by an Echo."""
+    return (
+        is_local(appliance)
+        and "CONTACT_SENSOR" in appliance["applianceTypes"]
```
alandtse: Is it possible appliance types could be None? If so, we should test for that to avoid key errors.

custom_components/alexa_media/binary_sensor.py
```diff
@@ -0,0 +1,109 @@
...
+    def unique_id(self):
+        return self.alexa_entity_id
+
+    @property
+    def is_on(self):
+        detection = parse_detection_state_from_coordinator(
+            self.coordinator, self.alexa_entity_id
+        )
+
+        return detection == 'DETECTED'
```
alandtse: Should we return none when the detection is none? This is the ha unavailable state used when there are connection issues. Right now, we default to false.

custom_components/alexa_media/binary_sensor.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @property
+    def assumed_state(self) -> bool:
+        last_refresh_success = (
+            self.coordinator.data and self.alexa_entity_id in self.coordinator.data
+        )
+        return not last_refresh_success
```
alandtse, line 109: This may have to factor in the unavailable comment above.

custom_components/alexa_media/const.py
```diff
@@ -30,6 +30,7 @@
     "sensor",
     "alarm_control_panel",
     "light",
+    "binary_sensor"
```
alandtse, line 33: I forget if the manifest.json or hacs.json may require and update for this too. Please check.

alandtse (changes requested):
Thanks for the PR. Please take a look at my comments.

custom_components/alexa_media/alexa_entity.py
cosimomeli: This function is copied from the one used in the light integration, and there's no check on applianceTypes there.

custom_components/alexa_media/binary_sensor.py
cosimomeli: assumed_state must return boolean and the documentation says """Return True if unable to access real state of the entity.""".

custom_components/alexa_media/binary_sensor.py
cosimomeli: Fixed. Now returns None if detection is None

custom_components/alexa_media/const.py
cosimomeli: There's no reference of the other constants if hacs.json or manifest.json, so I guess there's no need for binary_sensor too.

custom_components/alexa_media/alexa_entity.py
alandtse: Just because bad code made it in, doesn't mean we keep adding it. If you believe it won't be None, then that's fine. If you're wrong, it'll just break people's integration.

custom_components/alexa_media/binary_sensor.py
alandtse: Ok.

custom_components/alexa_media/alexa_entity.py
cosimomeli: I have no idea if applianceTypes could be None or not (and I don't know how to find this information), I just trusted the existing implementation. Anyway, as you are my guru here and there's no reason to argue about a null check, I just added a safe dict get of that key in both my function and in the existing `is_light` function.

alandtse: Great thanks.

alandtse (approved):

PR #1803, opened 2022-12-22
Add support for Amazon Indoor Air Quality Monitor

Snid0:
These changes will create 6 new sensors for each amazon air quality sensor detected. These get updated each time the component polls the alexa API. No updates are recieved on websockets.

Indoor Air Quality
Temperature 
Humidity
Particulate Matter
Carbon Monoxide
Volitile Organic Compounds

I pretty much copied the way the existing temperature sensor functionality worked. Expanding this to the extera sensors.


alandtse:
Thanks, I do recommend you learn how to use a git client so you don't have to use the GitHub website to add files. It will also make it easier for you to experiment.

custom_components/alexa_media/alexa_entity.py
```diff
@@ -188,6 +204,33 @@ def parse_alexa_entities(network_details: Optional[Dict[Text, Any]]) -> AlexaEnt
...
+                        serial if serial else appliance["entityId"]
+                    )                    
+                    # create array of air quality sensors. We must store the instance id against
+                    # the assetId so we know which sensors are which.
+                    sensors = []
+                    if appliance["friendlyDescription"] == "Amazon Indoor Air Quality Monitor" :
+                        for cap in appliance["capabilities"]:
+                            instance = cap.get("instance")
+                            if instance :                                
+                                friendlyName = cap["resources"].get("friendlyNames")
```
alandtse, line 231: Is it pretty clear the key `resources` will exist in `instance` exists? Want to avoid keyerrors for the majority of users who do not have IAQM.

custom_components/alexa_media/sensor.py
```diff
@@ -15,6 +15,7 @@
     DEVICE_CLASS_TIMESTAMP,
     STATE_UNAVAILABLE,
     TEMP_CELSIUS,
+    PERCENTAGE,
```
alandtse: Black should have sorted this alphabetically. Did you run black?

custom_components/alexa_media/sensor.py
```diff
@@ -226,6 +261,80 @@ def unique_id(self):
...
+        self._sensor_name = sensor_name
+        # tidy up name
+        self._sensor_name = self._sensor_name.replace("Alexa.AirQuality.",'')
+        self._sensor_name = ''.join(' ' + char if char.isupper() else char.strip() for char in self._sensor_name).strip()
+        self._name = name + " " + self._sensor_name
+        self._media_player_device_id = media_player_device_id
+        
+        self._instance = instance
+        
+        match unit:
```
alandtse: I normally like using dicts instead of match statements so the code is data driven by updating some values in const.py. What's your thoughts here?

alandtse (changes requested):
Thanks for this. I haven't tested it, but just have some questions for you. It doesn't necessarily mean make code changes but would like your thoughts.

custom_components/alexa_media/alexa_entity.py
Snid0: This code should only run if the appliance matches based on below.
if appliance["friendlyDescription"] == "Amazon Indoor Air Quality Monitor" :
I can only test against my device, but yes if instance existed there was always resources too.

custom_components/alexa_media/sensor.py
Snid0: Sorry total newb. I dont even know what black is? I added this manually. Didnt use anything to sort it.

custom_components/alexa_media/sensor.py
Snid0: I hadnt considered using dicts for this. It would certainly make the code tidier.

custom_components/alexa_media/alexa_entity.py
Snid0: Is there any neater way of traversing down the keys than using get? Is get the only safe way if you are unsure if the key is there?

custom_components/alexa_media/sensor.py
alandtse: Black is a style enforcer for python. I guess it's tied to the fact you're not using git, but the repo is set up to run a bunch of tests/enforcements before any git commit to enforce standards. I can run it for you before we accept the PR.

custom_components/alexa_media/sensor.py
alandtse: Do you mind making that change?

custom_components/alexa_media/alexa_entity.py
alandtse: I had missed it was nested in that if. That should be fine. But safest way is either nested gets or checking for keys before access.

Each level of nesting can be done like this with a get.
```python
something.get("key1", {}).get("key2", DEFAULT)
```

c3p0vsr2d2:
Works perfectly, thank you.
Two questions:
1. Is there any way to set the polling interval specifically for the air quality monitors? The sensor does report new values every minute, but seems that the integration polls data every 3 minutes?
2. I do not see new devices being created here, just additional entities - is this intended?

custom_components/alexa_media/sensor.py
alandtse: Nevermind, I handled it. Also I realized that match isn't available in python 3.9 which we still support.

custom_components/alexa_media/sensor.py
alandtse: Handled by precommit.

alandtse (approved):

Marcoletto:
HA see the entities but don't refresh them.
How can I solve?

Issue #1839, opened 2023-01-15
Please find a solution for Ledvance BT bulbs connected directly to an echo


Lippiero:
**NOTE: For a feature to be created it has to be doable in the Amazon app because we are basically replicating the app for HA. If you can't find it in the app, we probably can't create it here.**

**Is your feature request related to a problem? Please describe.**
My [bluetooth bulbs from Ledvance](
https://shop.ledvance.com/en/products/ledvance-smart-led-bluetooth-lamp-with-e27-socket-rgb-colour-changing-dimmable-replaces-60-w-bulb-warm-white-compatible-with-apple-homekit-and-ledvance-smart-app-for-android-energy-class-a) which are connected directly to an echo do not show up.
After reviewing the code, it seems that they are not detected as locally connected by alexa_entity.py#is_local. I checked the appliance's json and saw that connectedVia is not set (empty string), "ALEXA_VOICE_ENABLED" is not included in applianceTypes and the zigbee id pattern does not match (which is clear to me because it isn't a zigbee device):

> {
  "AAA_SonarCloudService_e42267783d4ef73e9f4a16c97ecc84a8": {
    "applianceId": "AAA_SonarCloudService_e42267783d4ef73e9f4a16c97ecc84a8",
    "endpointTypeId": "",
    "driverIdentity": {
      "namespace": "AAA",
      "identifier": "SonarCloudService"
    },
    "manufacturerName": "Ledvance",
    "friendlyDescription": "Ledvance intelligentes Gerät",
    "modelName": "",
    "deviceType": "",
    "version": "0",
    "friendlyName": "Erste Licht",
    "friendlyNameModifiedAt": 1673740898103,
    "capabilitiesModifiedAt": 1673740898103,
    "ipAddress": "",
    "port": "",
    "entityId": "41015137-bfda-4ae9-9c84-80594e38b690",
    "applianceNetworkState": {
      "reachability": "REACHABLE",
      "lastSeenAt": 1673740944739,
      "createdAt": 1673740898103,
      "lastSeenDiscoverySessionId": {
        "value": "5c3de69c-1ae1-4ef9-a813-f7f9d34e5905"
      }
    },
    "tags": {
      "tagNameToValueSetMap": {}
    },
    "additionalApplianceDetails": {
      "additionalApplianceDetails": {}
    },
    "firmwareVersion": "0",
    "actions": [],
    "capabilities": [
      {
        "capabilityType": "AVSInterfaceCapability",
[...]

PR #1847, opened 2023-01-29
fix: detect ledvance bluetooth bulbs #1839

Lippiero:
Draft for a change which let alexa media player detect local ledvance bulbs #1839

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -51,6 +51,11 @@ def is_hue_v1(appliance: dict[str, Any]) -> bool:
     return appliance.get("manufacturerName") == "Royal Philips Electronics"
 
 
+def is_skill(appliance: dict[str, Any]) -> bool:
+    namespace = appliance.get("driverIdentity", {}).get("namespace", "")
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. Isn't this a new feature? We never supported these lights before.

Lippiero:
I wasn't sure... it seems to be a bit of both... on the one hand it is a fix because the existing code catches other lights as well on the other hand yes, you are also right, if you say, they were not supported before. Should I change something?

I don't know, why one of the checks failed. There is no real hint given in the output. Is it ok though?

alandtse:
Don't worry about the pre-commit. For some reason it's not finding pylint.

Randyblo7 (approved):

Randyblo7 (commented):
Nice

Randyblo7 (approved):

Randyblo7 (approved):

Randyblo7 (approved):

Randyblo7 (approved):

Randyblo7 (approved):

Randyblo7 (approved):

Issue #1971, opened 2023-06-24
Notifications not updating automatically


francescolf:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/custom-components/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

When I set an alarm, timer, or reminder, the entities keep the unknown state.

**To Reproduce**

1. Start a timer
2. Check the state of `sensor.<echo name>_next_timer` or similar
3. The state is unknown

**Expected behavior**

The state should be the target time of the entity

**Screenshots**

None

**System details**

- Home-assistant (version): 2023.6.2
- alexa_media (version from `const.py` or HA startup): 4.6.4
- alexapy (version from `pip show alexapy` or HA startup): 1.26.8
- Amazon [2FA is enabled](https://github.com/custom-components/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). : y

**Logs**
Please provide [logs](https://github.com/custom-components/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

```
2023-06-24 22:26:56.550 DEBUG (MainThread) [custom_components.alexa_media] Finished fetching alexa_media data in 1.077 seconds (success: True)
2023-06-24 22:26:56.552 DEBUG (MainThread) [custom_components.alexa_media.media_player] c*********o@g*******m: <entity media_player.echo_di_francesco=standby> is last_called: G************2WE
2023-06-24 22:26:56.553 DEBUG (MainThread) [custom_components.alexa_media.media_player] c*********o@g*******m: Refreshing notify targets
2023-06-24 22:26:56.553 DEBUG (MainThread) [custom_components.alexa_media.notify] c*********o@g*******m: Found last_called <entity media_player.echo_di_francesco=standby> called at 1687638415022
2023-06-24 22:26:56.553 DEBUG (MainThread) [custom_components.alexa_media.notify] c*********o@g*******m: Creating last_called target last_called using <entity media_player.echo_di_francesco=standby> called at 1687638415022
2023-06-24 22:26:56.554 DEBUG (MainThread) [custom_components.alexa_media.notify] c*********o@g*******m: Found last_called <entity media_player.echo_di_francesco=standby> called at 1687638415022
[...]

PR #1974, opened 2023-06-28
Change sensor updating trigger to PUSH_ACTIVITY

francescolf:
Amazon stopped sending `NOTIFICATION_UPDATE` messages (at least for specific accounts, discussion in #1971)

This PR also adds a small delay in order to get updated data from the API.
These changes are proposed to the following accounts:
- amazon.it
- amazon.de
- amazon.com.au

The only current limitation is that Alexa commands must be sent ~10 seconds apart in order to update the sensors, but feedback is greatly appreciated.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -317,16 +317,7 @@ async def async_setup_entry(hass, config_entry):
         ),
     )
     hass.data[DATA_ALEXAMEDIA]["accounts"][email]["login_obj"] = login
-    no_nt_push_support = ["it", "de", "com.au"]
-    if any(d in login.url for d in no_nt_push_support):
-        hass.data[DATA_ALEXAMEDIA]["accounts"][email][
-            "notification_push_support"
-        ] = False
-        _LOGGER.debug("Amazon region does not support notification_push messages")
-    else:
-        hass.data[DATA_ALEXAMEDIA]["accounts"][email][
-            "notification_push_support"
-        ] = True
+    hass.data[DATA_ALEXAMEDIA]["accounts"][email]["last_push_activity"] = 0
     if not hass.data[DATA_ALEXAMEDIA]["accounts"][email]["second_account_index"]:
         hass.bus.async_listen_once(EVENT_HOMEASSISTANT_STOP, close_alexa_media)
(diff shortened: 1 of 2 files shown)


francescolf:
I replaced the command required to update the sensors since everyone should get _at least_ one `PUSH_ACTIVITY` message. However, for certain requests `PUSH_ACTIVITY` is received twice, triggering a useless double request which is now ignored.

alandtse:
Thanks. I think you could've left the legacy code in. Amazon sometimes reverts changes and even if it's not used now, it could be useful.

francescolf:
I'd prefer to keep it as is for two reasons:
1. The path for the serial number is different based on the message, resulting in more logic that increases latency
2. This solution triggers the update even if notification_update is supported since all the accounts receive push_activity messages

alandtse:
People don't really test so it's not something we have historically relied on. Your edits shouldn't break anything and solves broken behavior so I'll just push a release and people can roll back if there is an issue. Thanks!

dpgh947:
I just installed 4.6.5, doesn't work, timers still not updating. I would put this in the original issue but it's closed...

amaisano:
Timers going from <not set> to <timer> instantly get picked up. HOWEVER:

- Snooze events no longer update the time sensor
- Stopping a timer no longer changes the "Status" attribute from ON to OFF (it just stays on until HA restarts)

@alandtse

alandtse:
Don't tag people. It's rude. Amazon removed this feature. The fix is a stop gap but can only trigger on voice events. Any non voice interactions will not be picked up anymore.

amaisano:
Apologies.

The thing is, "Alexa, snooze timer" and "Alexa, stop time" are both "voice events," right? So I should expect those events to keep working... or am I misunderstanding something?

jane-t:
Any chance you could add the fix for the UK as well.  My timers are not working, or at least they feed in correctly when started but do no go away when stopped.

Randyblo7 (approved):

oneseventhree:
> Don't tag people. It's rude. Amazon removed this feature. The fix is a stop gap but can only trigger on voice events. Any non voice interactions will not be picked up anymore.

how an we go about doing this?

PR #2014, opened 2023-09-05
Add support to tts.cloud_say

surfingbytes:
Current Alexa notifitaction service doesn't support many languages, and doesn't support tts.cloud_day.
This PR adds support to tts.cloud_say by converting cloud generated mp3 via ffmpeg to format accepted by Alexa, and then playing that mp3. For this to work, HA instance has to be publicly accessible (either via Nabu Casa cloud subscription or any other custom solution), and for that reason, new setting was added to integration configuration.


alandtse:
Thanks. I can ignore the validate error but please fix the pre-commit requirements.

surfingbytes:
All the pre-commit issues are sorted, except of one. That is "comment", which is not comment, but string literal. But this is taken from original repo and is spread through the code, not just this place.

alandtse:
The issue is you put code above the docstring. Your code addition needs to be under it.

PR #2023, opened 2023-09-15
Update de.json

Ludy87:


alandtse:
Thanks. Please note this is auto updated via lokalise so should not be done via the jsons directly. https://github.com/custom-components/alexa_media_player/wiki/Translations

PR #2044, opened 2023-09-29
Restore NOTIFICATION_CHANGE

elarsson1:
This restores NOTIFICATION_CHANGE events for updating alarms and timers, which Amazon appears to support via HTTP2 notifications. This restores alarm and timer updates via changes in the Alexa app. I've kept PUSH_ACTIVITY as well as there does not seem to be any downside in responding to both event types and Amazon may be inconsistent in their implementation across different regions. 

Partial revert of 8f0dfa2 to restore NOTIFICATION_CHANGE functionality for timers and alarms.


alandtse:
Thanks!

PR #2071, opened 2023-10-13
Add switch and plug as "lights"

SRGDamia1:
This is a hacky fix that allows Alexa to find Bluetooth (and probably Zigbee) plugs and switches that are directly connected to Alexa.

Basically, I fake them as lights, and they are found using the existing machinery for external lights.


alandtse:
Thanks. We should really integrate them as switches or whatever makes sense. Treating them like lights will have users complain that the dimmer or colors don't work.

PR #2131, opened 2023-12-04
Update pt.json

ViPeR5000:


alandtse:
Thanks. Please note we do not accept PRs to the raw JSON files, we use this https://github.com/alandtse/alexa_media_player#translations. The json is overwritten automatically at the next release from what is in Lokalise

Issue #2149, opened 2023-12-28
Alexa Media Player media player not available in new beta


MJHeijster:
**Describe the bug**

Since upgrading to the new 2024.1 beta, the media_player entities are unavailable. The Do not disturb switch still is toggleable.

**To Reproduce**

Go to the integration and check the devices. Some media_player controls are unavailable, some are gone.

I have tried deleting and setting it up again, no result.

**System details**

- Home-assistant (version): **2024.1.0b0**
- alexa_media (version from `const.py` or HA startup): 4.8.1
- alexapy (version from `pip show alexapy` or HA startup): 1.27.10
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). : yes

**Logs**
Only thing off I could find was the following:
2023-12-28 16:26:23.362 ERROR (MainThread) [homeassistant.components.media_player] Error while setting up alexa_media platform for media_player
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 368, in _async_setup_platform
    await asyncio.gather(*pending)
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 507, in async_add_entities
    await asyncio.gather(*tasks)
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 752, in _async_add_entity
    await entity.add_to_platform_finish()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1275, in add_to_platform_finish
    self.async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 938, in async_write_ha_state
    self._async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1059, in _async_write_ha_state
    state, attr, capabilities, shadowed_attr = self.__async_calculate_state()
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 998, in __async_calculate_state
    attr.update(self.state_attributes or {})
                ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/media_player/__init__.py", line 1096, in state_attributes
    if self.support_grouping:
[...]

PR #2150, opened 2023-12-28
fix: fix media player supported features for HA beta

Petro31:
Fixes change in 2024.1 for supported features.

Fixes https://github.com/alandtse/alexa_media_player/issues/2149

### DIFF
--- a/custom_components/alexa_media/media_player.py
+++ b/custom_components/alexa_media/media_player.py
@@ -20,16 +20,11 @@ from homeassistant.components.media_player import (
     ATTR_MEDIA_ANNOUNCE,
     async_process_play_media_url,
 )
-from homeassistant.components.media_player.const import MEDIA_TYPE_MUSIC
 
 from homeassistant.const import (
     CONF_EMAIL,
     CONF_NAME,
     CONF_PASSWORD,
-    STATE_IDLE,
-    STATE_PAUSED,
-    STATE_PLAYING,
-    STATE_STANDBY,
     STATE_UNAVAILABLE,
 )
 from homeassistant.core import HomeAssistant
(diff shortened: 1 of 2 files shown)


custom_components/alexa_media/media_player.py
```diff
@@ -20,21 +20,8 @@
...
-    SUPPORT_PREVIOUS_TRACK,
-    SUPPORT_SELECT_SOURCE,
-    SUPPORT_SHUFFLE_SET,
-    SUPPORT_STOP,
-    SUPPORT_TURN_OFF,
-    SUPPORT_TURN_ON,
-    SUPPORT_VOLUME_MUTE,
-    SUPPORT_VOLUME_SET,
-)
+from homeassistant.components.media_player.const import MEDIA_TYPE_MUSIC
```
larena1: That is also deprecated

MJHeijster:
Thanks, hope it gets merged soon!

alandtse:
I guess we need to bumps hacs.json to set a new minimum version of HA.

alandtse (changes requested):
Please bump this line to cover the new HA version 2024.1. https://github.com/alandtse/alexa_media_player/blob/dev/hacs.json#L6. Also, can you check on the comment about the deprecation of `MEDIA_TYPE_MUSIC`. Ideally we can handle them all at the same time.

airy10:
Just confirming that it's fixing it for me, including the updating of sensors devices attached to a player

Petro31:
FYI, I didn't change STATE_UNAVAILABLE.  That's supposed to be handled by the available property only, not in the state.  I'll let you decide what to do with that.

When the state property returns None, the state will be translated to `unknown`.  When the availability property returns False, the state will translate to `unavailable`.

alandtse:
Thanks!

PR #2154, opened 2023-12-30
feat: add support for Amazon Smart Plug

alams154:
This feature allows the user to control Amazon Smart Plugs from the Alexa Media Player integration via switches in Home Assistant.

The feature is valuable since these devices cannot be added to Home Assistant and require complex workarounds.  For example, one implementation requires the use of various plugins used in conjunction (VoiceMonkey for routine triggers, Node-RED for state management and HTTP request methods, and Emulated Hue for state synchronization).

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -270,6 +270,7 @@ async def async_setup_entry(hass, config_entry):
                 "light": [],
                 "binary_sensor": [],
                 "temperature": [],
+                "smart_switch": [],
             },
             "entities": {
                 "media_player": {},
@@ -278,6 +279,7 @@ async def async_setup_entry(hass, config_entry):
                 "light": [],
                 "binary_sensor": [],
                 "alarm_control_panel": {},
+                "smart_switch": [],
             },
             "excluded": {},
             "new_devices": True,
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -127,6 +127,15 @@ def is_contact_sensor(appliance: dict[str, Any]) -> bool:
         and has_capability(appliance, "Alexa.ContactSensor", "detectionState")
     )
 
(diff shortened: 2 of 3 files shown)


alandtse:
Thanks. As a note this will be added after we raised minimum version to 2024.1.b0 so when I merge it you won't see it until you've updated.

alandtse (approved):

alams154:
Sounds good, thanks for all the hard work!

jvcsw:
Thanks, but how it works?

alams154:
> Thanks, but how it works?

image

Make sure you have "Include devices connected via Echo" enabled.  If your Amazon Smart Plug is correctly setup with your Amazon account on the Alexa app, the plug should automatically be discovered as a switch.  It will show up as a switch entity under the integration.

image

saberstop:
Hello, awesome work! At this point is the state of the plug expected to be working? I find that I am able to control the smart plug, but it won't actually update the state in Home Assistant if I use something other than HA like Alexa to turn on/off the plug. Thanks!

alams154:
Yep, I noticed that yesterday that the state wasn’t working.  I have some ideas on how to fix that which I’ll try implementing tomorrow.  I’m pretty new to Home Assistant development so I’m learning as I go.  I think the same problem exists with lights added by Alexa Media Player too, which is where I got my implementation from.

I think the main thing is the entity is missing an update/async_update function.  Depending on how the AlexaApy and Alexa handles state changes it might have a decent amount of delay in state changes.

I’m basing this off of the Wyze HACS integration which has about a 20 second delay from when you change it in Wyze and when you see the change in Home Assistant.

IceBotYT:
I know this is late, but thank you so much for this! I've just had this smart plug lying on my shelf for over a year now so to see it working finally is so satisfying to see. Can't wait to make more automations with it!!

NonaSuomy:
does this work with amazon light switches/dimmers as well?

Issue #2161, opened 2024-01-04
Alexa Guard warning within log on 2024.1.0 Home Assistant


MichaelStruck:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

Received a notice in my log files that says to open a bug report. Arming and Disarming of Alexa Guard continues to work, and I didn't see any other bug reports listing Alexa Guard:

Logger: homeassistant.components.alarm_control_panel
Source: helpers/deprecation.py:204
Integration: Alarm control panel (documentation, issues)
First occurred: 3:47:50 PM (80 occurrences)
Last logged: 9:02:01 PM

SUPPORT_ALARM_ARM_AWAY was used from alexa_media, this is a deprecated constant which will be removed in HA Core 2025.1. Use AlarmControlPanelEntityFeature.ARM_AWAY instead, please create a bug report at https://github.com/alandtse/alexa_media_player/issues

**To Reproduce**

Simply started receiving this after upgrading to 2024.1.0 HA and the latest 4.9.0 version of Alexa Media player. Everything else seems to be working fine.

**Expected behavior**

N/A

**Screenshots**

**System details**

- Home-assistant (version): 2024.1.0
- alexa_media (version from `const.py` or HA startup): 4.9.0
- alexapy (version from `pip show alexapy` or HA startup): ??
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). : Everything else is working great

**Logs**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

**Additional context**

figorr:
It looks that after the update there are also other errors related to deprecated features. I don’t have Alexa Guard in my region but I noticed an error related to some light features values.

> Logger: homeassistant.components.light
Source: components/light/__init__.py:1194
Integration: Luz (documentation, issues)
First occurred: 4 de enero de 2024, 19:51:11 (2 occurrences)
Last logged: 4 de enero de 2024, 19:51:11

[...]

PR #2182, opened 2024-02-07
fix: Fix deprecated alarm_control_panel constants

Snuffy2:
Fixes one part of #2161


alandtse:
Thanks. We may merge it in dev and wait a bit since it doesn't disappear until 2025.

PR #2183, opened 2024-02-11
fix: fix deprecated light constants and change to kelvin

Snuffy2:
# I have not fully tested this
I coded the changes below. Then when I went to test them, I found that the 1 of the lights that used to connect via Alexa was dead and I am unable to find where I put the other one (embarrassing). I don't normally use lights via Alexa, I just had them from some previous iterations of things I was working on.

So .... if people would like to test and report back, I can try to refine as needed. However, @alandtse, I'd totally understand if you just want to remove/close this PR since it is untested. Sorry.

This PR fixes the deprecated constants, but also changes to use kelvin behind the scenes since that is what the light entities now use and there isn't an easy way to replace the deprecated constants and keep using mireds.

### DIFF
--- a/custom_components/alexa_media/light.py
+++ b/custom_components/alexa_media/light.py
@@ -14,36 +14,18 @@ from typing import Optional
 from alexapy import AlexaAPI, hide_serial
 from homeassistant.components.light import (
     ATTR_BRIGHTNESS,
-    ATTR_COLOR_TEMP,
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. No worries; I don't have any lights and I merged PRs in blind before. I'd appreciate if you checked in on the repo after we release it to see if there's anything obvious to fix, but no worries either way.

Issue #2185, opened 2024-02-12
alexa connected lights no longer work


lsellens:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

After updating to v4.9.1 my son's echo glow is no longer available in home assistant. Says the entity is no longer provided. I tried removing and reinstalling the integration to no luck. Rolled back to v4.9.0 and his light was back.

![Screenshot at 2024-02-12 07-44-31](https://github.com/alandtse/alexa_media_player/assets/561979/b7d5c1b3-7229-4f7a-a96a-893551aa5405) 

**System details**

- Home-assistant (version): 2024.2.1
- alexa_media (version from `const.py` or HA startup): 4.9.1
- alexapy (version from `pip show alexapy` or HA startup): 1.27.10
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). : no

**Logs**

I can provide a full log if needed but I think this is the only relevant error.
```
Logger: homeassistant.components.light
Source: helpers/entity_platform.py:344
Integration: Light (documentation, issues)
First occurred: 7:42:17 AM (1 occurrences)
Last logged: 7:42:17 AM

Error while setting up alexa_media platform for light
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 344, in _async_setup_platform
    await asyncio.shield(task)
  File "/config/custom_components/alexa_media/light.py", line 98, in async_setup_entry
    return await async_setup_platform(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/light.py", line 78, in async_setup_platform
    light = AlexaLight(coordinator, account_dict["login_obj"], light_entity)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/light.py", line 135, in __init__
    self._attr_supported_color_modes = color_modes(details)
                                       ^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/light.py", line 118, in color_modes
[...]

PR #2186, opened 2024-02-12
fix: fix Color Mode Attribute

lsellens:
Fix: Replaced invalid attribute reference ColorMode.HS_COLOR with correct attribute ColorMode.HS
to address #2185


Snuffy2:
Thanks @lsellens, sorry about that.

PR #2195, opened 2024-02-26
Honor DND status

chemelli74:
Add "`dnd`" key in the notify service call to honor the DND status:

```yaml
service: notify.alexa_media_echo_dot_sala
data:
  message: test
  data:
    method: spoken
    type: announce
    dnd: true
```

If the new key is present and set to `true`, the message will not be played if the DND switch is active:

`2024-02-26 20:06:19.157 INFO (MainThread) [custom_components.alexa_media.media_player] DND active for device Echo Dot Sala, ignoring message`

In all other cases it will behave like before.

### DIFF
--- a/custom_components/alexa_media/media_player.py
+++ b/custom_components/alexa_media/media_player.py
@@ -1292,6 +1292,11 @@ class AlexaClient(MediaPlayerDevice, AlexaMedia):
     @_catch_login_errors
     async def async_send_announcement(self, message, **kwargs):
         """Send announcement to the media player."""
+        
+        if self.dnd_state and kwargs.get("dnd", False):
+            _LOGGER.info("DND active for device %s, ignoring message", self.name)
+            return
+        
         if self.hass:
             self.hass.async_create_task(
                 self.alexa_api.send_announcement(
--- a/custom_components/alexa_media/notify.py
+++ b/custom_components/alexa_media/notify.py
@@ -287,6 +287,7 @@ class AlexaNotificationService(BaseNotificationService):
                                 queue_delay=self.hass.data[DATA_ALEXAMEDIA]["accounts"][
                                     account
                                 ]["options"].get(CONF_QUEUE_DELAY, DEFAULT_QUEUE_DELAY),
+                                dnd=data["dnd"],
                             )
                         )
                         break


alandtse:
This honestly feels like an Amazon issue and not something we should be patching just for HA.

chemelli74:
Amazon says that DND is honoured only by their notification and not by playing music or messages.
Anyway I think that such enhancement doesn't hurt anyone as the default behaviour is not changed.

chemelli74:
**Status update**

I did a reset of Echo device a suggested by Alexa Support.
Now when I enable the DND from Alexa app I see a purple flash on the Echo device.
And if I play a message from HA, it's not played :rocket:

Only issue is that the DND of Alexa mobile App is not in sync with the DND switch in HA.

Can you please double check if this is the case for you as well ?

alandtse:
Does this need to be added for other options besides the announcement?

chemelli74:
> Does this need to be added for other options besides the announcement?

I think should be honoured for everything but alarms and reminders.

Did you get time to check if HA can be in sync with mobile app DND ?

alandtse:
No. I mean if you want this feature, I assume you'll do the research.

chemelli74:
Sorry a bit busy lately, will update you tomorrow

chemelli74:
Hi @alandtse thx for your patience.

Currently if you enable DND from Alexa Mobile App and then try to play a message from HA, it's not played as DND is honored.
No need to change anything from custom components code.

On the other side, the switch of this great custom components is OFF even if the DND on the Alexa Mobile App is ON.
Viceversa if you enable DND from the custom components HA switch, you find it OFF on Alexa Mobile App.

In the end we have  2 possible approaches:

- fix the DND switch status to be in sync with Alexa Mobile App ( preferred way)
- introduce the DND parameter in custom components

Let me know how we can work together to debug the issue and fix it.
Thank you again for your work and time!

chemelli74:
Hi mate, any idea on how to fix the sync between HA and Alexa app ?

alandtse:
Nope. I'm not actively developing this anymore. Just let me know when this is ready to merge though.

chemelli74:
> I'm not actively developing this anymore. 

You mean you are not developing this integration any more ?

alandtse:
Correct.

PR #2204, opened 2024-03-05
feat: Add support for non-Amazon switches and plugs

SRGDamia1:
This adds Alexa local switches as switches in addition to smart plugs. This also removes the requirement that local switches and plugs be Amazon-branded.  This allows Home Assistant to control things like [Sengled](https://www.amazon.com/Sengled-Plug-Compatible-Bluetooth-Required/dp/B093F6ZXYY) and other Bluetooth mesh plugs and switches directly connected to an Alexa device.

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -131,7 +131,10 @@ def is_switch(appliance: dict[str, Any]) -> bool:
     """Is the given appliance a switch controlled locally by an Echo."""
     return (
         is_local(appliance)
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. The concern was other brands may not work without testing. Did you at least confirm these changes work for the brands you mentioned?

SRGDamia1:
It works for my Sengled plugs (which register in Alexa as "switches") without compromising my Amazon plugs.

Those are the only two brands of Alexa-connected smart plugs I have.

PR #2241, opened 2024-05-01
Handle SMARTPLUG as lightswitch when defined in Alexa app as type "LIGHT"

BenGrimm42:
feat: handle SMARTPLUG as lightswitch when defined in Alexa app as type "LIGHT" (ie. SONOFF ZBMINI Extreme Zigbee switches or real smart plugs when used for lamps)


alandtse (approved):
Thanks!

Issue #2247, opened 2024-05-04
More deprecations affecting this integration.


Petro31:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**
```
2024-05-04 13:08:13.212 WARNING (MainThread) [homeassistant.helpers.frame] Detected that custom integration 'alexa_media' accesses hass.components.group. This is deprecated and will stop working in Home Assistant 2024.9, it should be updated to import functions used from group directly at custom_components/alexa_media/notify.py, line 240: entities.extend(self.hass.components.group.expand_entity_ids(entities)), please create a bug report at https://github.com/alandtse/alexa_media_player/issues
```
line 240 in notify.py

```
entities.extend(self.hass.components.group.expand_entity_ids(entities))
```
I'll see if I can find out what is required to change this line.

**To Reproduce**

Run a tts notification.

**Expected behavior**

To not use the deprecated API

**Screenshots**

**System details**

- Home-assistant (version): 2024.5.x
- alexa_media (version from `const.py` or HA startup): 4.9.2
- alexapy (version from `pip show alexapy` or HA startup): whatever version 4.9.2 uses in the manifest.
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). : y

**Logs**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).
```
2024-05-04 13:08:13.212 WARNING (MainThread) [homeassistant.helpers.frame] Detected that custom integration 'alexa_media' accesses hass.components.group. This is deprecated and will stop working in Home Assistant 2024.9, it should be updated to import functions used from group directly at custom_components/alexa_media/notify.py, line 240: entities.extend(self.hass.components.group.expand_entity_ids(entities)), please create a bug report at https://github.com/alandtse/alexa_media_player/issues
```
**Additional context**

Petro31:
probably can just change to

```
entities.extend(expand_entity_ids(self.hass, entities))
```

PR #2248, opened 2024-05-04
fix: fix hass.component.group deprecation.

Petro31:
Fixes https://github.com/alandtse/alexa_media_player/issues/2247.


alandtse (approved):
Thanks, do we need to bump the min ha version to access that function?

Petro31:
That's a good question, I'll look into it in the morning.

Petro31:
Yes, the change to the function was made in 2024.2.0b0, so I assume this was added then.

https://github.com/home-assistant/core/pull/106924

alandtse:
Can you also bump the minimum ha in hacs.json please?

Issue #2225, opened 2024-04-05
Deprecation warning custom integration alexa_media calls async_add_job


fwmarcel:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

Home Assistant issues the following warning

> Detected that custom integration 'alexa_media' calls `async_add_job`, which is deprecated and will be removed in Home Assistant 2025.4; Please review https://
> developers.home-assistant.io/blog/2024/03/13/deprecate_add_run_job for replacement options at custom_components/alexa_media/__init__.py, line 625: hass.async_add_job

**System details**

- Home-assistant (version): 2024.4.0
- alexa_media (version from `const.py` or HA startup): v4.9.2

micium:
The message is also reported for media_player.py:

Detected that custom integration 'alexa_media' calls `async_add_job`, which is deprecated and will be removed in Home Assistant 2025.4; Please review https://developers.home-assistant.io/blog/2024/03/13/deprecate_add_run_job for replacement options at custom_components/alexa_media/media_player.py, line 188: hass.async_add_job(, please create a bug report at https://github.com/alandtse/alexa_media_player/issues

danielgoepp:
Same issue, but I didn't find this right away due to the lack of details in the subject. Who can edit subject here to make it a bit more descriptive and spelled correctly? Perhaps something like: 

Deprecation warning custom integration alexa_media calls async_add_job

This will avoid confusion with other related issues like #2229

hcassus:
Does it make sense to only change `custom_components/alexa_media/media_player.py:188` to `async_create_task` and we'd already be fine?

Marcoletto:
I could think that we will find a new firmware with that problem fixed?

'''
Registratore: homeassistant.helpers.frame
Fonte: helpers/frame.py:188
Prima occorrenza: 12 aprile 2024, 20:56:14 (4 occorrenze)
Ultima registrazione: 12 aprile 2024, 20:58:46

Detected that custom integration 'hacs' accesses hass.components.frontend. This is deprecated and will stop working in Home Assistant 2024.9, it should be updated to import functions used from frontend directly at custom_components/hacs/frontend.py, line 68: hass.components.frontend.async_register_built_in_panel(, please create a bug report at https://github.com/hacs/integration/issues
[...]

PR #2250, opened 2024-05-05
fix: fix deprecation of async_add_job

Petro31:
Fixes deprecation warnings for async_add_job.

Fixes https://github.com/alandtse/alexa_media_player/issues/2225

partially fixes: https://github.com/alandtse/alexa_media_player/issues/2239


Petro31:
I don't believe this will change the min version if we merge the other deprecation warnings.

PR #2251, opened 2024-05-05
build: bump min HA version in hacs.json

Petro31:


alandtse:
Oh. Thanks. I meant to include it in the PR which requires the new minimum. But I think this works too.

PR #2260, opened 2024-05-10
fix: fix lights being identified as switches (#2204)

BenGrimm42:
fix: commit e3da18b to support switches as lights (if redeclared that way in the Alexa app) was incomplete due to conflicting change in commit 2344bcf (#2204)

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -131,13 +131,14 @@ def is_contact_sensor(appliance: dict[str, Any]) -> bool:
     )
 
 def is_switch(appliance: dict[str, Any]) -> bool:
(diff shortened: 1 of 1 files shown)


BenGrimm42:
Sorry! Needed to fix my previous change again.
When I made the intial change for commit e3da18b I did (and tested) that based on the 4.9.2 release version and not on the current dev branch. And the commit 2344bcf (#2204), which was already on the dev branch, unfortunately makes my change useless, as it now handled all switches from all manufacturers and thus also "stealing" the ones that are redeclared as lights.
The new change now excludes the "lights" from the is_switch handler, so that they get again through to the later happening is_light handler.

BenGrimm42:
But at least that didn't break anything in 4.10, the "light" switches where just again not found as lights, but just as switches.

alandtse (approved):

PR #2294, opened 2024-06-29
fix: fix warning for async_forward_entry_setup

Petro31:
Fix warning about awaiting `hass.config_entries.async_forward_entry_setup`.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -624,8 +624,8 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
                     _LOGGER.debug("Loading config entry for %s", component)
                     config_entry.async_create_task(
                         hass,
-                        hass.config_entries.async_forward_entry_setup(
-                            config_entry, component
+                        hass.config_entries.async_forward_entry_setups(
+                            config_entry, [component]
                         )
                     )
                 else:
--- a/custom_components/alexa_media/media_player.py
+++ b/custom_components/alexa_media/media_player.py
@@ -187,8 +187,8 @@ async def async_setup_entry(hass, config_entry, async_add_devices):
                 )
                 config_entry.async_create_task(
                     hass,
-                    hass.config_entries.async_forward_entry_setup(
-                        config_entry, component
+                    hass.config_entries.async_forward_entry_setups(
+                        config_entry, [component]
                     )
                 )
         return True
--- a/custom_components/alexa_media/notify.py
+++ b/custom_components/alexa_media/notify.py
@@ -20,6 +20,7 @@ from homeassistant.components.notify import (
     BaseNotificationService,
 )
 from homeassistant.const import CONF_EMAIL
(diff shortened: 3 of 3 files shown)


bdraco:
The call should be awaited instead of creating a task as it means the lock can be released before setup is complete otherwise

Petro31:
> The call should be awaited instead of creating a task as it means the lock can be released before setup is complete otherwise 

I originally had it that way but I also noticed there were no warnings for scheduling the task.  I'll put up another PR to just await the func instead of creating the task.

danielbrunt57:
Is this the correct method then?
```
#                    config_entry.async_create_task(
#                        hass,
#                        hass.config_entries.async_forward_entry_setups(
#                            config_entry, [component]
#                        )
#                    )
                    await hass.config_entries.async_forward_entry_setups(
                            config_entry, [component]
                        )
```
I just tried it and it worked, but it didn't solve the partitioned cookie issue, as you pointed out.  I think we'll have to wait until we see HA update to python 3.13 near the end of the year.  I poked inside alexapy a little to see if there was a way to override http cookies.py with a newer version but was unsuccessful.  And I highly doubt HA would entertain the thought of updating cookies.py in the current 3.12  with the one which we won't see until October at the earliest, just because of AMP's issue..

On a somewhat related note, I've been recoding this integration to make all options reconfigurable. I'm very new to this low level stuff and it's taken me a while to get it working successfully (a lot of learning involved on my part).  I don't have a dev system so have been working with my live system (AMP 4.10.3) night and day and many MANY restarts!  It's now time to present it but I'm unsure of the best way.  Just update my fork so it's got my current code and submit a PR and see where it goes?  Thanks.

danielbrunt57:
How do I solicit help in updating the 14 non-English translation files???

alandtse:
Really shouldn't comment on closed threads. It'll be missed in most projects. I highly recommend using a dev container for dev. That just needs the main ha repo. It should work out of the box. You'll need a script to copy the files into the container, but that should be easy or maybe it's automatic now.

Yes do a PR. It'd be better if each thing was separate. Like await in it's own, etc.

For translations, we need to upload the new strings.json and then ask people to help in lokalise. See wiki. It normally takes a few releases before everything gets translated.

danielbrunt57:
Thanks Alan.  I knew this wasn't the right door to be knocking on but it was the only door I could see at this early hour after have spent all night finishing this (and all week working on it)!
The last little tidbit was getting hass_url to save to core.config entries so it could be reconfigured.  I had noticed `self.config.pop("hass_url")` before but I finally clued in only hours ago that's what was preventing me from saving and was  raising an error on my reconfig as invalid key.  In the original version, it's not reconfigurable and thus no longer needs to be retrieved.  Which leads me to a nagging question: with all your knowledge, can it be reconfigured safely in an existing config entry?  I was thinking of skipping it as I don't know the repercussions, plus, it's not likely going to be needed, but hey, someday, someone might.  Also, it not really a big deal to re & re.  I think I have that down to under 15 seconds now!

image

PR #2327, opened 2024-07-18
fix: en.json somehow reverted to the old version

danielbrunt57:
For some reason, en.json was still the old version and the configure options screen was missing all descriptions except one.


alandtse:
I told you it has to be done in lokalise.

PR #2328, opened 2024-07-18
fix: en.json is the old version

danielbrunt57:
The configure options window is missing all descriptions except one.


danielbrunt57:
It was Edge browser - deleted browsing data and problem solved.

Issue #2338, opened 2024-07-20
AttributeError: 'AlexaMediaFlowHandler' object has no attribute 'config_entry'


danielbrunt57:
**Describe the bug**

 ```
File "/config/custom_components/alexa_media/config_flow.py", line 808, in _update_schema_defaults
    default=self.config_entry.options.get(CONF_QUEUE_DELAY, 1.5),
            ^^^^^^^^^^^^^^^^^
AttributeError: 'AlexaMediaFlowHandler' object has no attribute 'config_entry'
```
**To Reproduce**

1. Restart Home Assistant
2. Alexa Media Player fails to initialize
3. Select 'Reload'

**Expected behavior**

There should be no error and integration should reload

**Screenshots**

**System details**

- Home-assistant (version): 2024.7.4
- alexa_media (version from `const.py` or HA startup): 4.11.1
- alexapy (version from `pip show alexapy` or HA startup): 1.27.10
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n) : Y

**Logs**
```
Logger: homeassistant.config_entries
Source: config_entries.py:586
First occurred: 2:55:35 PM (4 occurrences)
Last logged: 2:57:57 PM

Error setting up entry daniel@brunt.ca - amazon.com for alexa_media
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 586, in async_setup
    result = await component.async_setup_entry(hass, self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/__init__.py", line 348, in async_setup_entry
    if await test_login_status(hass, config_entry, login):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/__init__.py", line 1395, in test_login_status
    ] = await hass.config_entries.flow.async_init(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 1267, in async_init
    flow, result = await self._async_init(flow_id, handler, context, data)
[...]

PR #2339, opened 2024-07-20
fix: Update config_flow.py

danielbrunt57:
Line 808 should be: default=self.config.get(CONF_QUEUE_DELAY, 1.5),

Tested reload after failure and it's fixed.


danielbrunt57:
This resolves Issue #2338

danielbrunt57:
I missed line 812!

alandtse:
Please stop doing this open close pr thing. Just use the same PR.

Issue #2333, opened 2024-07-20
500 Internal Server Error Server got itself in trouble


Pico1965:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

At the end of the configuration, after entering the capcha the error appears:
500 Internal Server Error 
Server got itself in trouble

**To Reproduce**

1. Go to Settings
2. Click on Devices & Services
3. Scroll down to Alexa Media Player
4. Configure until capcha solved

**System details**

HA:
Core 2024.7.3
Supervisor 2024.06.2
Operating System 12.4
Frontend 20240710.0
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). Yes

Integration:
alexa_media_player: 4.11.**1**

**Logs**
```
2024-07-20 12:23:52.086 ERROR (MainThread) [aiohttp.server] Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/httpx/_transports/default.py", line 69, in map_httpcore_exceptions
    yield
  File "/usr/local/lib/python3.12/site-packages/httpx/_transports/default.py", line 373, in handle_async_request
    resp = await self._pool.handle_async_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/httpcore/_async/connection_pool.py", line 216, in handle_async_request
    raise exc from None
  File "/usr/local/lib/python3.12/site-packages/httpcore/_async/connection_pool.py", line 196, in handle_async_request
    response = await connection.handle_async_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/httpcore/_async/connection.py", line 101, in handle_async_request
    return await self._connection.handle_async_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/httpcore/_async/http11.py", line 143, in handle_async_request
    raise exc
[...]

PR #2343, opened 2024-07-21
fix: use `config_flow` in OptionsFlow Handler

danielbrunt57:
Three fixes/changes:
```
    def async_get_options_flow(
        config_entry: config_entries.ConfigEntry,
    ) -> config_entries.OptionsFlow:
```
```
    def __init__(self, config_entry: config_entries.ConfigEntry) -> None:
```
and `default=self.config_entry.data.get()` for all options.

### DIFF
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -830,9 +830,6 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
         return OptionsFlowHandler(config_entry)
 
 
-######
-
-
 class OptionsFlowHandler(config_entries.OptionsFlow):
     """Handle a option flow for Alexa Media."""
 
@@ -865,16 +862,16 @@ class OptionsFlowHandler(config_entries.OptionsFlow):
                 (
                     vol.Optional(
                         CONF_SCAN_INTERVAL,
-                        default=self.config_entry.data.get(CONF_SCAN_INTERVAL, 120),
+                        default=self.config_entry.data.get(
+                            CONF_SCAN_INTERVAL, 120
+                        ),
                     ),
                     int,
                 ),
                 (
                     vol.Optional(
                         CONF_QUEUE_DELAY,
-                        default=self.config_entry.data.get(
-                            CONF_QUEUE_DELAY, DEFAULT_QUEUE_DELAY
-                        ),
+                        default=self.config_entry.data.get(CONF_QUEUE_DELAY, DEFAULT_QUEUE_DELAY),
                     ),
                     float,
                 ),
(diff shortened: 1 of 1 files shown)


danielbrunt57:
Resolves latest issue in #2333

danielbrunt57:
I think I now need to register an update listener to the config entry that will be called when the entry is updated?

custom_components/alexa_media/config_flow.py
```diff
@@ -823,11 +823,16 @@ def _update_schema_defaults(self) -> Any:
...
     @callback
-    def async_get_options_flow(config_entry):
+    def async_get_options_flow(
+        config_entry: config_entries.ConfigEntry,
+    ) -> config_entries.OptionsFlow:
         """Get the options flow for this handler."""
         return OptionsFlowHandler(config_entry)
 
 
+######
```
alandtse: What is this?

custom_components/alexa_media/config_flow.py
```diff
@@ -911,6 +917,11 @@ async def async_step_init(
             self.hass.config_entries.async_update_entry(
                 self.config_entry, data=user_input, options=self.config_entry.options
             )
+
+            self.hass.config_entries.async_update_entry(
```
alandtse, line 915: Why twice?

alandtse (changes requested):

custom_components/alexa_media/config_flow.py
danielbrunt57: Sorry!

custom_components/alexa_media/config_flow.py
danielbrunt57: An error!  Fixing...

danielbrunt57: BTW, is it HA convention or your personal styling preference to have two blank lines preceding classes/functions as I'm noticing quite a bit of that??

alandtse: It's a style enforced by black. I don't waste time arguing style. It may have come from ha or whatever but no one in python has time to argue over format.

alandtse (approved):

custom_components/alexa_media/config_flow.py
```diff
@@ -908,6 +911,7 @@ async def async_step_init(
                 user_input[CONF_OTPSECRET] = self.config_entry.data[CONF_OTPSECRET]
             if CONF_OAUTH in self.config_entry.data:
                 user_input[CONF_OAUTH] = self.config_entry.data[CONF_OAUTH]
+
```
alandtse, line 914: Not sure why this space was added but if it helps with your understanding it's fine.

alandtse:
Btw timeouts are probably ipv6. Which has been answered so much. But whatever, I'm ok with closing the later reported issues in that issue.

PR #2349, opened 2024-07-21
chore: Update translations

AndryGamingYT:


AndryGamingYT:
I updated the translation files `it.json` and `en.json` with the missing strings. I also reordered the various strings in the same order as the `strings.json` file.

alandtse:
Thanks, but please note we use lokalise to handle translations. Please make edits there.

AndryGamingYT:
Hi, I updated those files because I noticed a bug in the UI not showing translations. In fact, there were checkboxes and nothing was written next to them, or input fields without placeholders.
Adding what was missing solved the problem.
I checked the project on Lokalise and noticed that they are indeed missing there too. In particular, some strings are missing in options::step::init::data which should be added.

alandtse:
Thanks. Again, those json files are built whenever there is a release and sourced from lokalise. So as long as it's fixed in lokalise it'll be fixed automatically.

AndryGamingYT:
Can you upload the updated `strings.json` file to Lokalise so you can proceed with translations?

alandtse:
It was uploaded already when added to git. Are the keys missing

AndryGamingYT:
Yes, there are some missing keys. I will write down what they are:
- `options::step::init::data::hass_url`
- `options::step::init::data::include_devices`
- `options::step::init::data::exclude_devices`
- `options::step::init::data::scan_interval`
- `options::step::init::data::debug`
- `options::step::init::description`
- `options::step::init::title`

alandtse:
Thanks for catching that. I reuploaded it and it found more keys.

PR #2353, opened 2024-07-22
fix: fix async request refresh

danielbrunt57:
hass.data[DATA_ALEXAMEDIA]["accounts"].setdefault() "options": {}
Changed config_entry.options.get() to config_entry.data.get()

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -304,12 +304,8 @@ async def async_setup_entry(hass, config_entry):
             "second_account_index": 0,
             "should_get_network": True,
             "options": {
-                CONF_INCLUDE_DEVICES: config_entry.data.get(
-                    CONF_INCLUDE_DEVICES, ""
-                ),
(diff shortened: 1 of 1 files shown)


custom_components/alexa_media/__init__.py
```diff
@@ -1345,19 +1355,23 @@ async def update_listener(hass, config_entry):
...
             _LOGGER.debug(
-                "Changing option %s from %s to %s",
+                "Option %s changed from %s to %s",
                 key,
                 old_value,
                 hass.data[DATA_ALEXAMEDIA]["accounts"][email]["options"][key],
             )
-            if key == CONF_EXTENDED_ENTITY_DISCOVERY:
-                reload_needed = True
+            #            if key == CONF_EXTENDED_ENTITY_DISCOVERY:
```
alandtse: Don't leave dead code in. if you don't think we need it, just remove it.

danielbrunt57: Sorry, I missed that one after debugging/testing.

Issue #2306, opened 2024-07-08
non thread-safe operation


tbclark3:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

Every day or two I get a runtime error related to non thread-safe operation.  Sometimes HA continues and sometimes it becomes unresponsive, spewing out repeated errors. I have update interval set to 10 minutes, but I'm not sure if this error occurs during an attempted update from Amazon.  It occurs whether or not an automation has recently sent audio to the speakers.

**To Reproduce**

Run Alexa media player integration.

**Expected behavior**

I expect Alexa media player to run without errors.

**Screenshots**

**System details**

- Home-assistant (version): 2024.7.1
- alexa_media (version from `const.py` or HA startup): 4.10.3
- alexapy (version from `pip show alexapy` or HA startup):
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). : yes

**Logs**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).
Logger: homeassistant
Source: helpers/frame.py:197
First occurred: 11:19:55 AM (1 occurrences)
Last logged: 11:19:55 AM

Error doing job: Future exception was never retrieved (None)
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/sensor.py", line 433, in _trigger_event
    self.hass.bus.async_fire(
  File "/usr/src/homeassistant/homeassistant/core.py", line 1505, in async_fire
    frame.report_non_thread_safe_operation("hass.bus.async_fire")
  File "/usr/src/homeassistant/homeassistant/helpers/frame.py", line 225, in report_non_thread_safe_operation
    report(
  File "/usr/src/homeassistant/homeassistant/helpers/frame.py", line 159, in report
    _report_integration(what, integration_frame, level, error_if_integration)
[...]

Issue #2264, opened 2024-05-12
'alexa_media' calls async_fire from a thread at custom_components/alexa_media/sensor.py


Marcoletto:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

Registratore: homeassistant.helpers.frame
Fonte: helpers/frame.py:190
Prima occorrenza: 13:35:17 (1 occorrenze)
Ultima registrazione: 13:35:17

Detected that custom integration 'alexa_media' calls async_fire from a thread at custom_components/alexa_media/sensor.py, line 433: self.hass.bus.async_fire(, please report it to the author of the 'alexa_media' custom integration

**System details**

- Home-assistant (version):
Core
2024.5.3
Supervisor
2024.05.1
Operating System
12.3
Frontend
20240501.1

Alexa Media Player 4.10.0 (last update yesterday)

Marcoletto:
Registratore: homeassistant
Fonte: helpers/frame.py:203
Prima occorrenza: 20:03:00 (2 occorrenze)
Ultima registrazione: 20:50:00

Error doing job: Future exception was never retrieved
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/sensor.py", line 433, in _trigger_event
    self.hass.bus.async_fire(
  File "/usr/src/homeassistant/homeassistant/core.py", line 1508, in async_fire
    self._hass.verify_event_loop_thread("async_fire")
  File "/usr/src/homeassistant/homeassistant/core.py", line 440, in verify_event_loop_thread
    frame.report(
  File "/usr/src/homeassistant/homeassistant/helpers/frame.py", line 162, in report
    _report_integration(what, integration_frame, level, error_if_integration)
  File "/usr/src/homeassistant/homeassistant/helpers/frame.py", line 203, in _report_integration
    raise RuntimeError(
RuntimeError: Detected that custom integration 'alexa_media' calls async_fire from a thread at custom_components/alexa_media/sensor.py, line 433: self.hass.bus.async_fire(. Please report it to the author of the 'alexa_media' custom integration.

[...]

PR #2357, opened 2024-07-24
fix: add callback decorator to hide threading warning

danielbrunt57:
Fix issues #2306 and #2264 - error `alexa_media' calls hass.bus.async_fire from a thread other than the event loop` when alarms are dismissed or snoozed.

### DIFF
--- a/custom_components/alexa_media/sensor.py
+++ b/custom_components/alexa_media/sensor.py
@@ -423,6 +423,7 @@ class AlexaMediaNotificationSensor(SensorEntity):
                     dt.as_utc(self._attr_native_value),
                 )
 
+    @callback
     def _trigger_event(self, time_date) -> None:
         _LOGGER.debug(
             "%s:Firing %s at %s",


danielbrunt57:
Alan, I'm not sure this is the correct fix as I can't tell if `def _process_raw_notifications(self):` is running in the event loop or the executor as I don't really understand how all that is done code-wise.  I understand the concept of threading but when I look at the code I'm not sure what's running where..

alandtse:
Actually, this just lies to HA and says there's no synchronous calls. I don't think this is a true solution although it'll shut up the warning until they add another test.

danielbrunt57:
Can I go ahead and merge it then?

alandtse:
I don't think this is a solution. You're hiding the message by declaring a callback. It's not fixing it.

danielbrunt57:
I had previously messaged Petro over in HACF and got his thoughts about `@callback` and `self.hass.bus.fire` vs `self.hass.bus.async_fire`.  Rather than duplicating it all here, can I add you into the conversation?

alandtse:
I mean if you're positive this is the fix, then I'll trust you but last I checked `@callback` was just a decorator that doesn't change the underlying function.

alandtse:
And honestly, any serious code discussions should be public so people understand why it was made.

danielbrunt57:
I posted a topic in HACF but got zero responses.  I then messaged Petro and he responded.  His answer was in two parts:
[quote="petro, post:3, topic:753137, full:true"]
The callback one is more correct. I'm not 100% sure what callback does when going through the event bus, but it does hold special meaning. It basically adds a property to the coroutine that the event bus does something specific with.
[/quote]
[quote]
The async_track_point_in_utc_time does that. Callback just adds a property to the function.
```
def callback[_CallableT: Callable[..., Any]](func: _CallableT) -> _CallableT:
    """Annotation to mark method as safe to call from within the event loop."""
    setattr(func, "_hass_callback", True)
    return func
```
I just checked the code, it allows the coroutine to be flagged as a HassJob, which lets HA know it’s safe to call within the event loop.
[/quote]

I'm not sure how else to go public with it?

I tested dismissing an alarm both with it running in executor code using `self.hass.bus.async_fire` and again with it running in the event loop via `self.hass.bus.async_fire` decorated with `@callback` and both solutions worked without an error.  If you'd like to keep it in executor code rather than event loop, I'm fine with that.  I've found other integrations with the error and one fixed it with `@callback` and `self.hass.bus.async_fire` and another fixed it with `hass.bus.fire`.  I don't really know if one is more correct than the other...

alandtse:
Honestly, we lived without the warning for years before HA found a reason to detect it. I guess we can go a few more years.

danielbrunt57:
So @callback is kind of like saying "Trust me, this routine is perfectly safe so go ahead and put it in the event loop!"??

alandtse:
Yes.

alandtse:
At least that's how I understood it. It's been years since I checked it's implementation.

Petro31:
> I posted a topic in HACF but got zero responses. I then messaged Petro and he responded. His answer was in two parts: [quote="petro, post:3, topic:753137, full:true"] The callback one is more correct. I'm not 100% sure what callback does when going through the event bus, but it does hold special meaning. It basically adds a property to the coroutine that the event bus does something specific with. [/quote] [quote] The async_track_point_in_utc_time does that. Callback just adds a property to the function.
> 
> ```
> def callback[_CallableT: Callable[..., Any]](func: _CallableT) -> _CallableT:
>     """Annotation to mark method as safe to call from within the event loop."""
>     setattr(func, "_hass_callback", True)
>     return func
> ```
> 
> I just checked the code, it allows the coroutine to be flagged as a HassJob, which lets HA know it’s safe to call within the event loop. [/quote]
> 
> I'm not sure how else to go public with it?
> 
> I tested dismissing an alarm both with it running in executor code using `self.hass.bus.async_fire` and again with it running in the event loop via `self.hass.bus.async_fire` decorated with `@callback` and both solutions worked without an error. If you'd like to keep it in executor code rather than event loop, I'm fine with that. I've found other integrations with the error and one fixed it with `@callback` and `self.hass.bus.async_fire` and another fixed it with `hass.bus.fire`. I don't really know if one is more correct than the other...

I was under the impression that you moved to `hass.bus.fire` while marking it as a `callback`, if that's not the case, then this PR is not correct.  `has.bus.fire` is the correct path.

PR #2359, opened 2024-07-25
feat: add timer label to Timer sensor attributes

Canis-L-Sapien:
This change adds the name of a timer to the timer entity attribute list. This is similar to how reminder titles show up on Reminder entities.


alandtse:
Thanks!

PR #2364, opened 2024-07-28
chore: update manifest.json requirements alexapy==1.28.1

danielbrunt57:


alandtse:
alexpy bumps require more changes including pyproject and the devcontainer. Take a look at past bumps. Please note poetry.lock requires running `poetry update`. It cannot be updated by hand.

danielbrunt57:
> alexpy bumps require more changes including pyproject and the devcontainer.

So this has to me done by you then in alexapy?

alandtse:
You can do them here but you need to run poetry update to update the lock file. Look at the other bumps in the history.

PR #2367, opened 2024-07-28
fix: apply correct fix for threading warning

danielbrunt57:
Remove `@callback`
Change `hass.bus.async_fire` to `hass.bus.fire`


alandtse:
As a note a revert must be atomic and only revert the prior commit. Since you're doing more, it's just a fix.

Issue #2378, opened 2024-08-01
Integration can't load with 2024.8.0b0


ChristophCaina:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

The integration can't load with HomeAssistant 2024.8.0b0
The following error will be logged:

```
Logger: homeassistant.config_entries
Quelle: config_entries.py:604
Erstmals aufgetreten: 07:00:02 (2 Vorkommnisse)
Zuletzt protokolliert: 07:00:29

Error setting up entry c*******h@c******e - amazon.de for alexa_media
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 604, in async_setup
    result = await component.async_setup_entry(hass, self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/alexa_media/__init__.py", line 353, in async_setup_entry
    await login.login(cookies=await login.load_cookie())
  File "/usr/local/lib/python3.12/site-packages/alexapy/helpers.py", line 137, in wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alexapy/alexalogin.py", line 643, in login
    resp = await self._session.get(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/aiohttp/client.py", line 481, in _request
    raise RuntimeError("Session is closed")
RuntimeError: Session is closed
```

**To Reproduce**

**Expected behavior**

![image](https://github.com/user-attachments/assets/32064d56-cc09-4583-9f31-a6d33cb9e3ae)

**System details**

- Home-assistant (version):
- alexa_media (version from `const.py` or HA startup):
- alexapy (version from `pip show alexapy` or HA startup):
- Amazon [2FA is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n). :

**Logs**
```
[...]

PR #2380, opened 2024-08-01
fix: use async_create for notification

ChristophCaina:
fixed issue with partition cookies
(https://gitlab.com/keatontaylor/alexapy/-/commit/e594a3b194eec7d6451ad0c7135d8c32704f5f73)

unfortunately, this change does not solve the issue #2378

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -38,6 +38,7 @@ from homeassistant.const import (
     EVENT_HOMEASSISTANT_STOP,
 )
 from homeassistant.core import callback
+from homeassistant.components.persistent_notification import async_creat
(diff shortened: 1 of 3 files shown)


alandtse:
Thanks. However a dependency bump requires changes in multiple places. Please take a look at past bumps to see what is required.

ChristophCaina:
Thanks,
I will Check this.

ChristophCaina:
should now also fix #2382 - but since my installation is not loading, I cannot test this further :-(

alandtse:
Since the alexapy bump isn't complete, let's just do the persistent notification with this PR.

ChristophCaina:
reverted the version bump back :)

PR #2413, opened 2024-08-07
chore: Update poetry.lock to update to alexapi 1.28.1 with "poetry update"

jleinenbach:
Used "poetry update" to create an updated poetry.lock file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,14 +50,14 @@ repos:
   - hooks:
       - id: black
     repo: https://github.com/psf/black
-    rev: 23.11.0
+    rev: 24.8.0
   - repo: https://github.com/pre-commit/mirrors-prettier
     hooks:
       - id: prettier
         exclude: ^custom_components/alexa_media/translations|CHANGELOG.md
     rev: v4.0.0-alpha.8
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.15.0
+    rev: v3.17.0
     hooks:
       - id: pyupgrade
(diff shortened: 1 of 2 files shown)


chrisvblemos:
I think a single PR for bumping alexapy, with all necessary changes (devcontainer, poetry...), would be better, no? At least this is the case for previous bump PRs.

alandtse:
Yes. Please bump all in a single.

danielbrunt57:
"Stop the presses", if you can!
My merged PR for `alexapy` to skip "partitioned" key on loading the cookie was not always 100% successful and I'm currently testing a fix which skips the partitioned key when _saving_ the session cookie to the jar.  I've only tested it with a current config entry (which fails initialization due to a partitioned cookie). I need to test it on a new config entry as well as see if it survives a reauthentication/token refresh.  The fix I'm testing in alexapy v1.27.10 is as simple as, and in lieu of, my previous one:
`alexalogin.py` v1.27.10
lines 1253 (changed) -1254 (unchanged)
```
if name in ["Name", "Value"] or name == "partitioned":`
    continue
```

jleinenbach:
@danielbrunt57 
Done. Do you think we should bump alexapy after this is fixed, too?
https://gitlab.com/keatontaylor/alexapy/-/issues/32

danielbrunt57:
> @danielbrunt57 Done.

I've completed testing with setting up a fresh config entry and it does not work with this edit in place due to a new `captcha error` so I am abandoning this idea.  Please proceed as previously planned...

PR #2417, opened 2024-08-07
fix: fix hass_url undefined

### LINKED ISSUE #2408: config_flow.py line 815: `hass_url` undefined
Sorry everyone but I just had an automatic re-auth situation which triggered the `update_schema_defaults `and unfortunately there was a pre-existing `self.config.pop("hass_url")` at line 621 in `async def _test_login` making `hass_url `unavailable after that.

I removed lines 621-622, restarted and the problem was resolved.

Issuing a new PR for it...

danielbrunt57:
Removed self.config.pop("hass_url")
Fixes #2408

### DIFF
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -617,8 +617,6 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
                 self.config.pop("reauth")
             if self.config.get(CONF_SECURITYCODE):
                 self.config.pop(CONF_SECURITYCODE)
-            if self.config.get("hass_url"):
-                self.config.pop("hass_url")
             self.config[CONF_OAUTH] = {
                 "access_token": login.access_token,
                 "refresh_token": login.refresh_token,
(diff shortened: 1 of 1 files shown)


LorenzoRogai (approved):

chrisvblemos:
@danielbrunt57 Could you also include the following change to this PR as described by @LorenzoRogai [here](https://github.com/alandtse/alexa_media_player/issues/2408#issuecomment-2274440500)? 

Can't get more undefined than that.

chrisvblemos:
I'm not sure yet why this pop was there to begin with and can't comment on whether removing it could break other things. I think the root issue is at line 815. Solution is very simple, as given by @LorenzoRogai, and should be included anyway even if the .pop is unnecessary as it is a critical error (hass_url doesn't exist in the given context):

Change
`CONF_HASS_URL, default=self.config.get(CONF_HASS_URL, hass_url)`
to:
`CONF_HASS_URL, default=self.config.get(CONF_HASS_URL, DEFAULT_HASS_URL)`

danielbrunt57:
I've compared v4.12.5 with v4.10.3 and in 4.10.3, `hass_url `was never in `_update_schema_defaults`:
```
                vol.Required(
                    CONF_HASS_URL, default=self.config.get(CONF_HASS_URL, hass_url)
                ): str,
```
I think I added it with the intention it be a reconfigurable option but after a little discussion with Alan, decided to not make it reconfigurable and that section of code lingered.  I've tested my `config_flow` leaving lines 620-621 with `config.pop` in `_test_login` as they were previously but removed lines 814-816 in `_update_schema_defaults`.  Initial configuration worked and also a reconfigure options so I believe at this point, that is the correct fix.

grantclem:
Sadly not working for .com.au

PR #2423, opened 2024-08-08
fix: .de login dont work

jressel01:
fix .de login dont work


ChristophCaina:
the last time, when I wanted to update the file, I got the feedback, that also some other files needs to be updated.
I don't really know why - since it seems to be an update of the HASH...

It's probably for the built in checks here...

saschaabraham:
The commit message should be "fix: .de login dont work". The : is missing.

ChristophCaina:
> The commit message should be "fix: .de login dont work". The : is missing.

it's not only `.de` domain...
According to the feedback in the issue, several other domains are affected as well.

PR #2461, opened 2024-08-17
feat: change extended entity discovery initial default from False to True

danielbrunt57:
Change DEFAULT_EXTENDED_ENTITY_DISCOVERY from False to True.
This should help avoid future issues when users forget they need to turn this on to see temperature, AQM & other attached sensors!


alandtse:
No. Users should enable options. There are users who handle connected devices via other integrations. This would be a breaking change for them.

alandtse:
We also get enough complaints that we generate too many entities. This would make that worse.

PR #2476, opened 2024-08-20
fix: fix deprecated hass.components use

danielbrunt57:
config_flow.py: Use async_dismiss_persistent_notification


danielbrunt57:
Before you release a new version, can we get the alexapy version in manfest.json bumped to 1.29.0?

Issue #2477, opened 2024-08-20
Included device (comma separated)/Excluded device (comma separated) are not working as expected


danielbrunt57:
**Describe the bug**

Exclude Devices is excluding more than just the device I want excluded. 
I.e. specifying "Echo Auto" excludes not only "Echo Auto" but also "Echo" since device name "Echo" is in the _string_ "Bedroom,Daniel's Fire,Echo Auto,Mach-E SYNC,Living Room,Main floor,Norton Court,Office,This Device"

**To Reproduce**

Amazon device names  with the issue:  "Echo" and "Echo Auto"
Exclude Devices _string_: "Bedroom,Daniel's Fire,Echo Auto,Mach-E SYNC,Living Room,Main floor,Norton Court,Office,This Device"
I tried `Echo Auto`, `Echo%20Auto` and `Echo\ Auto` but none of these worked.

**Expected behavior**

Only exclude device "Echo Auto" but not device "Echo"

**System details**

- Home-assistant (version): 2024.8.2
- alexa_media (version from `const.py` or HA startup): 4.12.10
- alexapy (version from `pip show alexapy` or HA startup): 1.29.0
- Amazon [2FA/2SV is enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (y/n): Y

**Logs**
> 024-08-20 00:13:30.345 DEBUG (MainThread) [custom_components.alexa_media] d****l@b******a: Existing: [] New: ['Garage Echo Dot', 'Living Room Echo Dot Right', 'Office Echo Dot Left', 'Office Echo Dot Right', 'Bedroom Echo Dot Left', 'Kitchen Echo Dot', 'Bedroom Echo Dot Right', 'Echo Show 8', 'Living Room Echo Dot Left']; Filtered out by not being in include: [] or in exclude: ['Echo', 'Echo Auto', 'Office', 'Main floor', 'Mach-E SYNC', 'Bedroom', 'Norton Court', 'Living Room', "Daniel's Fire", 'This Device']

MrPiket:
Something that I also notice in those fields when you do the configuration via the UI is that you cannot leave that field empty. When you click on submit, it doesn't delete the value of that field it goes back with what you have before.

PR #2478, opened 2024-08-20
fix: process include/exclude devices

danielbrunt57:
Process `include_devices` and `exclude_devices` strings as CSV lists.
Fixes issue #2477

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -537,7 +537,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         for device in devices:
             serial = device["serialNumber"]
             dev_name = device["accountName"]
-            if include and dev_name not in include:
+            if include and dev_name not in include.splt(','):
                 include_filter.append(dev_name)
                 if "appDeviceList" in device:
                     for app in device["appDeviceList"]:
@@ -550,7 +550,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
                     serial
                 ] = device
                 continue
-            if exclude and dev_name in exclude:
+            if exclude and dev_name in exclude.splt(','):
                 exclude_filter.append(dev_name)
                 if "appDeviceList" in device:
                     for app in device["appDeviceList"]:
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -537,7 +537,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         for device in devices:
             serial = device["serialNumber"]
             dev_name = device["accountName"]
-            if include and dev_name not in include.splt(','):
+            if include and dev_name not in include.splt(","):
                 include_filter.append(dev_name)
                 if "appDeviceList" in device:
                     for app in device["appDeviceList"]:
@@ -550,7 +550,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
                     serial
                 ] = device
                 continue
-            if exclude and dev_name in exclude.splt(','):
+            if exclude and dev_name in exclude.splt(","):
                 exclude_filter.append(dev_name)
                 if "appDeviceList" in device:
                     for app in device["appDeviceList"]:
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -537,7 +537,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         for device in devices:
             serial = device["serialNumber"]
             dev_name = device["accountName"]
-            if include and dev_name not in include.splt(","):
+            if include and dev_name not in include:
                 include_filter.append(dev_name)
                 if "appDeviceList" in device:
                     for app in device["appDeviceList"]:
@@ -550,7 +550,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
                     serial
                 ] = device
                 continue
-            if exclude and dev_name in exclude.splt(","):
+            if exclude and dev_name in exclude:
                 exclude_filter.append(dev_name)
                 if "appDeviceList" in device:
                     for app in device["appDeviceList"]:
@@ -1243,9 +1243,18 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
     _LOGGER.debug("Setting up Alexa devices for %s", hide_email(login_obj.email))
     config = config_entry.data
     email = config.get(CONF_EMAIL)
-    include = config.get(CONF_INCLUDE_DEVICES)
-    exclude = config.get(CONF_EXCLUDE_DEVICES)
-    scan_interval: float = (
+    include = (
+        cv.ensure_list_csv(config[CONF_INCLUDE_DEVICES])
+        if config[CONF_INCLUDE_DEVICES]
+        else ""
+    )
+    _LOGGER.debug("include: %s", include)
+    exclude = (
+        cv.ensure_list_csv(config[CONF_EXCLUDE_DEVICES])
+        if config[CONF_EXCLUDE_DEVICES]
+        else ""
+    )
+    _LOGGER.debug("exclude: %s", exclude)    scan_interval: float = (
         config.get(CONF_SCAN_INTERVAL).total_seconds()
         if isinstance(config.get(CONF_SCAN_INTERVAL), timedelta)
         else config.get(CONF_SCAN_INTERVAL)
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -1254,7 +1254,8 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         if config[CONF_EXCLUDE_DEVICES]
         else ""
     )
-    _LOGGER.debug("exclude: %s", exclude)    scan_interval: float = (
+    _LOGGER.debug("exclude: %s", exclude)
+    scan_interval: float = (
         config.get(CONF_SCAN_INTERVAL).total_seconds()
         if isinstance(config.get(CONF_SCAN_INTERVAL), timedelta)
         else config.get(CONF_SCAN_INTERVAL)


alandtse:
Something has changed. Those should already be lists. Not strings. The ingestion step should already convert them from strings.

jleinenbach:
I've noticed that the code uses `.splt` instead of `.split()` in both changes:

```python
if include and dev_name not in include.splt(","):
```

and

```python
if exclude and dev_name in exclude.splt(","):
```

Just to clarify, is this intentional or could it be a typo, possibly due to a copy/paste error? I'm asking because `.splt` is not a standard Python method, and I'm wondering if this was meant to be `.split()`.

Thanks!

jleinenbach:
### Problem Analysis by ChatGPT

The issue you’re encountering relates to the logic used for handling the inclusion and exclusion of devices based on their names. Specifically, the problem arises when a device name partially matches a string in the `exclude` list. This can result in unintended exclusions, where devices with shorter names (e.g., "Echo") are excluded when the intent was to exclude a more specific device (e.g., "Echo Auto").

### Root Cause

Let's break down the exact problem with an example:

1. **Partial String Matching:**  
   Consider the device names `"Echo"` and `"Echo Auto"`. If the `exclude` list contains `"Echo Auto"`, but the exclusion logic uses a partial string match (e.g., using `in` operator), then `"Echo"` may inadvertently be excluded because `"Echo"` is a substring of `"Echo Auto"`.

   **Code Example:**
   ```python
   exclude = "Echo Auto"
   dev_name = "Echo"
   
   if dev_name in exclude:
       # This will wrongly exclude "Echo"
       continue
   ```
   Here, `dev_name in exclude` is `True` because `"Echo"` is found within the string `"Echo Auto"`.

2. **Handling of Whitespace within Names:**  
   Device names like `"Echo Auto"` have spaces in them. If the exclusion list is treated as a single string and split by commas, any device name that partially matches an element in the `exclude` list could be wrongly excluded. Additionally, if a device name has leading or trailing spaces, or inconsistent internal spacing, it might not be correctly matched or excluded.

3. **Incorrect Assumption about List Handling:**  
   There might be an assumption that `include` and `exclude` are already lists, but if these are indeed strings, they must be properly split and cleaned (trimmed of whitespace) before use. If this is not done, the logic might not work as intended, especially with names that contain spaces.

### Detailed Example

- **Given Devices:**
  - `"Echo"`
  - `"Echo Auto"`
  
- **Exclude List:** `"Bedroom,Daniel's Fire,Echo Auto,Mach-E SYNC,Living Room,Main floor,Norton Court,Office,This Device"`

If the exclusion logic checks if `"Echo"` is in the `exclude` list as a substring (i.e., using `in` operator), it will incorrectly match and exclude `"Echo"`, because `"Echo"` is a part of `"Echo Auto"`. The logic fails to distinguish between complete names and partial matches within a string.

### Expected Behavior

The correct behavior should ensure that `"Echo Auto"` is excluded, but `"Echo"` is not, as they are distinct devices. The exclusion logic should perform an exact match rather than a partial match.

### Proposed Solution

To address these issues, the following adjustments should be made to the code:

1. **Ensure Proper List Conversion:**
   Convert `exclude` and `include` from strings to lists if they are not already lists. This involves splitting the string by commas and trimming any extraneous whitespace.

2. **Exact Matching Instead of Partial Matching:**
   Use exact matching rather than the `in` operator for checking whether a device name should be excluded or included.

3. **Whitespace Handling:**
   Normalize device names and elements in the `exclude` list by stripping leading and trailing spaces and ensuring consistent internal spacing.

### Implemented Solution

```python
# Ensure exclude and include are lists
if isinstance(exclude, str):
    exclude = [item.strip() for item in exclude.split(",")]

if isinstance(include, str):
    include = [item.strip() for item in include.split(",")]

# Normalize the device name by stripping spaces
dev_name_normalized = dev_name.strip()

# Priority for the exclude list: if the device name is in exclude, skip it
if dev_name_normalized in exclude:
    continue

# Include the device only if it's explicitly listed in include (if include is not empty)
if include and dev_name_normalized not in include:
    continue

# At this point, the device is included if it's in the include list and not in the exclude list
```

### Why This Works

- **Exact Matching:** This solution ensures that only exact matches in the `exclude` list lead to exclusion, which prevents partial matches from causing unintended exclusions.
  
- **Whitespace Handling:** By trimming whitespace, we avoid issues where names with accidental spaces (either leading, trailing, or within the name) cause mismatches.

- **List Handling:** Converting the `exclude` and `include` strings into properly handled lists ensures the logic is applied correctly across all device names.

jleinenbach:
> Something has changed. Those should already be lists. Not strings. The ingestion step should already convert them from strings.

With the assistance of ChatGPT, I believe I've identified the root cause of the issue:

**Potential Issue in `config_flow.py`:**

There is a potential problem in the way the `CONF_INCLUDE_DEVICES` and `CONF_EXCLUDE_DEVICES` lists are handled within the `_save_user_input_to_config` method. Specifically, the use of `reduce` to convert these lists into comma-separated strings can lead to unintended behavior.

### Explanation:
- **Use of `reduce`:** When the `CONF_INCLUDE_DEVICES` or `CONF_EXCLUDE_DEVICES` lists are provided as input, the current code uses `reduce` to concatenate the list items into a single string, separated by commas:
  ```python
  if isinstance(user_input[CONF_INCLUDE_DEVICES], list):
      self.config[CONF_INCLUDE_DEVICES] = (
          reduce(lambda x, y: f"{x},{y}", user_input[CONF_INCLUDE_DEVICES])
          if user_input[CONF_INCLUDE_DEVICES]
          else ""
      )
  ```
  This transformation turns a list like `["Echo Auto", "Bedroom"]` into the string `"Echo Auto,Bedroom"`. While this might seem convenient, it introduces a risk where the string may be incorrectly processed later, particularly in matching logic.

- **Implication:** Once converted to a string, any subsequent matching logic could mistakenly identify partial matches as valid exclusions or inclusions. For example, if `CONF_EXCLUDE_DEVICES` contains `"Echo Auto"`, it might be incorrectly treated as `"Echo,Auto"`, leading to unintended exclusions like `"Echo"`.

### Recommendation:
Avoid converting lists into comma-separated strings using `reduce`. Instead, maintain the list structure to ensure that each device name is matched exactly as intended. This approach will prevent issues where partial names are incorrectly excluded or included due to improper string processing.

chrisvblemos (changes requested):
There is a typo in split(). But I wouldn't apply the fix here, since in this case you would change the expected type to string instead of a list, which goes against the rest of the code (i.e, the function you changed expects an Optional[list[str]], not a Optional[str]. If you think this is the proper way, I'd change the rest of the code to conform to this change. 

Instead, I'd just apply the fix inside _save_user_input_to_config() at config_flow.py so that instead of transforming config[CONF_...] insto a str, it keeps it as a list.

alandtse:
So this is probably worth checking as these were the prior assumptions:

1. Strings are inputs from the config entry fields and also from the `.storage` location. Only code processing the config info needs to treat them as strings and will help convert to Lists.
2. Service calls probably also have to treat inputs as strings but also convert to lists.
3. All strings should have been converted to lists once live in HA. All interaction should be with lists within the component.

If any of these have been broken, then they need to be fixed. This also suggests the proposed fix here is also incorrect since it breaks 3. Let's plan to close this after a few days unless your investigation shows that assumptions are wrong.

danielbrunt57:
Sorry, I've been busy today with 8 hours of landscaping work to make some $$$, and my body is not used to physical labor any more!

@jleinenbach Yes, it's a stupid typo but not due to copy/paste error, instead a failure to use copy/paste and attempt to type such a small change correctly, and fail!

@alandtse, @chrisvblemos  
This is not a new issue from recent code changes that I'm aware of, rather it's a recent change in how I configure AMP that's brought it to light.  After dozens and dozens of re & re's and setting areas for my 18 devices, I paused for reflection on why I needed some of them at all in HA and so started using `exclude_devices` thus uncovering what's likely been this way for I don't know how long.  

I read the comments this morning and went back though AMP's version history until I found mention of UI config and that was v2.0.0.  I downloaded it and saw this in config_flow.py:
```
Line 140
```
    async def async_step_user`
```
Lines 163-178:
```
        if isinstance(user_input[CONF_INCLUDE_DEVICES], str):
            self.config[CONF_INCLUDE_DEVICES] = (
                user_input[CONF_INCLUDE_DEVICES].split(",")
                if "CONF_INCLUDE_DEVICES" in user_input
                else []
            )
        else:
            self.config[CONF_INCLUDE_DEVICES] = user_input[CONF_INCLUDE_DEVICES]
        if isinstance(user_input[CONF_EXCLUDE_DEVICES], str):
            self.config[CONF_EXCLUDE_DEVICES] = (
                user_input[CONF_EXCLUDE_DEVICES].split(",")
                if "CONF_EXCLUDE_DEVICES" in user_input
                else []
            )
        else:
            self.config[CONF_EXCLUDE_DEVICES] = user_input[CONF_EXCLUDE_DEVICES]
```
So they used to be stored back then as lists in the .storage location.

I will re-review more recent code versions to pinpoint when it changed.
Right now, I'm suspicious that I may have done it with #2319...

**Comment by danielbrunt57** [comment]
> 3. All strings should have been converted to lists once live in HA. All interaction should be with lists within the component.

I'm not seeing that in 4.12.8 **_or 4.10.3_**, which is before my Options feat change so I'm sure now I did not break it and think this has been a longstanding unnoticed possible failure point, which my particular setup has just now revealed.  Or it may have been encountered before but never reported...

In `__init__.py`, `async def setup_alexa` inputs them as strings (since they are saved in .storage as strings):
```
1266    include = config.get(CONF_INCLUDE_DEVICES)
1267    exclude = config.get(CONF_EXCLUDE_DEVICES)
```
 `include` & `exclude` are only used in `__init__.py` in lines 540 & 553 where I made my change.

So if I am to obey point #3, then it looks like the fix should be applied at lines 1266,1267?
However, if I use `exclude = config.get(CONF_EXCLUDE_DEVICES).split(',')` then media_player is broken.  Why? I don't currently know but with this code and I get no alexa media players.  I also get zero devices in AMP, jjust73 entities.
Therefore,  I either have to fix it:
a) the way I did.
or
b) leave include/exclude as strings, set `exclude_list = exclude.splt(',')` and change `if exclude and dev_name in exclude` to `if exclude and dev_name in exclude_list` (and similarly for include)

Is `b)` the preferred way to do it?

**Comment by alandtse** [comment]
Make sure you understand fully how exclude/include [work ](https://github.com/alandtse/alexa_media_player/wiki/Configuration#inclusions)as that confuses people. 

That said, make sure that the behavior with empty lists is sane. For include, it used to ignore an empty list. If you've modified the code and it actually tries to process an empty list, then nothing will be included since they're not explicitly allowed. Exclude should not have that issue at all as it's a ban list and an empty list shouldn't ban anything.

**Comment by danielbrunt57** [comment]
@alandtse Yes, I understand how exclude/include [work](https://github.com/alandtse/alexa_media_player/wiki/Configuration#inclusions).
I will test any new code modifications to ensure empty lists remain functional.

> @alandtse For include, it used to ignore an empty list. If you've modified the code and it actually tries to process an empty list, then nothing will be included since they're not explicitly allowed. Exclude should not have that issue at all as it's a ban list and an empty list shouldn't ban anything.

Debugging tells me `config.get(CONF_INCLUDE_DEVICES)` returns a string.
I do not see any code to convert the `include` and `exclude` strings to lists which would allow `if include and dev_name not in include` to correcty build the `include_filter[]` and `exclude_filter[]` lists.

> @chrisvblemos Instead, I'd just apply the fix inside _save_user_input_to_config() at config_flow.py so that instead of transforming config[CONF_...] into a str, it keeps it as a list.

I was thinking of that but that would seem to go against Alan's three  assumptions.

My original (rejected) fix works and after spending another 5 hours on this, the only other way I see to solve it is:
```
    _LOGGER.debug("Setting up Alexa devices for %s", hide_email(login_obj.email))
    config = config_entry.data
    email = config.get(CONF_EMAIL)

    include = config[CONF_INCLUDE_DEVICES]
    if include:
        include = include.split(',')
    _LOGGER.debug("include devices: %s", include)

    exclude = config[CONF_EXCLUDE_DEVICES]
    if exclude:
        exclude = exclude.split(',')
    _LOGGER.debug("exclude devices: %s", exclude)

    scan_interval: float = (
```

**Comment by danielbrunt57** [comment]
@chrisvblemos 
I tried changing the transformation from list to str inside `_save_user_input_to_config(self, user_input=None)` like this:
```
        if CONF_EXCLUDE_DEVICES in user_input:
            if isinstance(user_input[CONF_EXCLUDE_DEVICES], str):
                self.config[CONF_EXCLUDE_DEVICES] = (
                    user_input[CONF_EXCLUDE_DEVICES].split(',')
                    if user_input[CONF_EXCLUDE_DEVICES]
                    else []
                )
            else:
                self.config[CONF_EXCLUDE_DEVICES] = user_input[CONF_EXCLUDE_DEVICES]
```
but it continues to save the config entry in `.storage as `"exclude_devices": "Bedroom,Daniel's Fire,Echo Auto,Mach-E SYNC,Living Room,Main floor,Norton Court,Office,This Device",`.  Either this code `user_input[CONF_EXCLUDE_DEVICES].split(',')`is incorrect, or it's something else.

I don't know why the integration changed from lists to strings for these parameters sometime between v2.0.0 and now.
The only place I see INCLUDE & EXCLUDE being used is at lines 1246-1247 & 537-565 in `__init__.py` to build its include and exclude filters to filter the list of potential devices.

I'm going to leave it you guys to figure out the best fix.  Once you do, I can amend my PR with your recommended changes or, you can submit a new PR to fix my issue and I'll close this one.  But I am done trying to wrap my my head around everything for now.  Maybe in 3-6 months after I've readjusted my life, I can resume doing what I love doing...

alandtse:
Thanks again for all you help. Good luck with your IRL stuff!

PR #2502, opened 2024-09-03
fix: delete cookiefile only with `remove_entry`

danielbrunt57:
Add `async_remove_entry` and move cookiefile deletion from `async_unload_entry` to `async_remove_entry`.


alandtse:
Where are these functions called? It seems weird that the fix for deleting too much is to write a new function `remove_entry` function. Is this a new HA function? If so, please provide a link to the dev docs or the github showing it was added. This may also require a HA version bump depending on when it was introduced.

alandtse:
Nevermind, saw it in the linked issue.

danielbrunt57:
I wish I could figure out how to terminate the http2 connection when the entry is removed.

The closest I can see is the `close_callback=http2_close_handler` but it's purpose seems to be to try and reconnect it when it closes rather than terminate it, and the `await login_obj.close()` doesn't end the HTTP2EchoClient process.  It persists even after the config entry is removed...

PR #2518, opened 2024-09-12
feat: add model name resolution

lymanepp:
I found a long list of model id to model names, but it has at least some errors, so I only enabled those that I was able to verify.

Screenshot_20240912-014334


alandtse:
Awesome. Thanks for this. I think we should just enable them all and let people identify when wrong. Otherwise, only devs who happen to look in the code would know they could comment it in to see if it works. Do you mind uncommenting those entries?

lymanepp:
I enabled all model name translations and added a comment in the code

alandtse (approved):

Issue #2579, opened 2024-10-03
Version v4.13.3 is making HA fail


CancunManny:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

I installed v4.13.3 and shortly after my HA kept loosing connection.  I would re boot, it would work fine for a few minutes, then it would disconect.

**To Reproduce**

I used a full backup for HA for Sept 30th.  Before the Alexa update I did a core update, assumed that caused the issue.  Once I restored I didn't do the core update but did do the Alexa update.  Few minutes later issue of disconnect started again.

Restored the Sept 30th backup again. This time I did the HA core update, but skipped the v4.13.3 update for Alexa.  My HA has been stable for over 30 minutes now. 

**Expected behavior**

I would expect to install v4.13.3 and not have my HA system crash.

**Screenshots**

No screen shots

**System details**

- Home Assistant version: Running HA via proxmox server.  Updated to the last version
- alexa_media version (from `const.py` or HA startup log): Stable version 4.13.2
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log):
- Is Amazon [2FA/2SV enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account)  (y/n):
- Amazon Domain:

**Debug Logs (alexa_media & alexapy)**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

**Additional context**

synologyy:
yes this updates make my system ERROR dont install this update!!!

JackTalisker:
I have the same big issue.
HA goes to 100% CPU and begins to eat alla the RAM.
System unusable after few seconds. No time di downgrade AMP.

![Screenshot 2024-10-03 083906](https://github.com/user-attachments/assets/cd1d6dbd-c075-494a-a544-b3479356a701)

ocean90:
Having the same issue! Had to reboot in safe mode (which took a few trials) and deactivate the integration.

It seems like then installing the old version makes HA stable again.

LeoeLeoeL:
OUT OF MEMORY error and HA reboots every 5 minutes.
Now I disabled the integration; I hope is enough.

[...]

PR #2580, opened 2024-10-03
fix: Fixes issue #2579

jleinenbach:
### Summary of the CPU Issue and Solution

**Problem:**  
The implementation with **parallel loading** of Alexa components caused high CPU usage. Multiple components were loaded simultaneously, which overwhelmed the system due to the concurrent resource-heavy operations.

**Solution:**
1. **Serial Loading:**   I addressed the CPU problem by **serializing** the loading process. This means components are now loaded **one after the other** instead of all at once, reducing the CPU load.
   
2. **Condition with `entry_setup`:**   I reintroduced the check for `if not entry_setup`. This ensures that we only initialize components if they are not already set up, preventing redundant operations. For components already initialized, the same `async_forward_entry_setup` method is called to reload them.

**Outcome:**  
By loading components sequentially and ensuring that unnecessary setups are skipped, we significantly reduce the CPU load while maintaining efficient processing of new Alexa clients.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -1538,7 +1538,7 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
                     await hass.config_entries.flow.async_abort(flow_id)
                     _LOGGER.debug("Successfully aborted config flow '%s' for %s", flow_id, hide_email(email))
                 except UnknownFlow:
-                    _LOGGER.warning("Config flow '%s' not found for %s", flow_id, hide_email(email)))
+                    _LOGGER.warning("Config flow '%s' not found for %s", flow_id, hide_email(email))
                 except Exception as ex:
                     _LOGGER.error("Error aborting config flow '%s' for %s: %s", flow_id, hide_email(email), ex)
             # Remove the flow from the config_flows dictionary
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -356,13 +356,16 @@ async def async_setup_entry(hass, config_entry):
     if not hass.data[DATA_ALEXAMEDIA]["accounts"][email]["second_account_index"]:
         hass.bus.async_listen_once(EVENT_HOMEASSISTANT_STOP, close_alexa_media)
         hass.bus.async_listen_once(EVENT_HOMEASSISTANT_STARTED, complete_startup)
-    listener_relogin_required = hass.bus.async_listen("alexa_media_relogin_required", relogin)
-    listener_relogin_success = hass.bus.async_listen("alexa_media_relogin_success", login_success)
+    listener_relogin_required = hass.bus.async_listen(
+        "alexa_media_relogin_required", relogin
+    )
+    listener_relogin_success = hass.bus.async_listen(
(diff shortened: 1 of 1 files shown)


MichaelMKKelly (changes requested):
line 1541 needs a ) removed from the end to fix a syntax error

custom_components/alexa_media/__init__.py
```diff
@@ -1343,63 +1430,171 @@ async def http2_error_handler(message):
...
-        _LOGGER.debug("Removing config_flows data")
-        async_dismiss_persistent_notification(
-            hass, f"alexa_media_{slugify(email)}{slugify((entry.data['url'])[7:])}"
-        )
-        hass.data[DATA_ALEXAMEDIA].pop("config_flows")
-    if not hass.data[DATA_ALEXAMEDIA]:
-        _LOGGER.debug("Removing alexa_media data structure")
-        if hass.data.get(DATA_ALEXAMEDIA):
-            hass.data.pop(DATA_ALEXAMEDIA)
+                    _LOGGER.warning("Config flow '%s' not found for %s", flow_id, hide_email(email)))
```
apbarratt:
```suggestion
                    _LOGGER.warning("Config flow '%s' not found for %s", flow_id, hide_email(email))
```

apbarratt: As flagged by @MichaelMKKelly

alandtse:
Please note, while I'm not opposed to what looks like ChatGPT changes, given the recent issue with #2579, I'm going to require someone actually test it before merging.

That said, I want to thank you for being willing to do PRs.

danielbrunt57:
@jleinenbach Are you aware of [PR #2589](https://github.com/alandtse/alexa_media_player/pull/2580) to address memory exhaustion/infinite loop?

danielbrunt57:
@alandtse As you're no doubt aware, I implemented @jleinenbach's original code and then amended the PR with 3 edits to address issues I encountered with it.  That said, I'm running the resulting `__init__.py` in my system and everything here is working fine.  
Also, I just now implemented the `helpers.py` code change in PR [#2589](https://github.com/alandtse/alexa_media_player/pull/2580) and everything is still working.

dpwood:
The problem as described by this PR doesn't fit reports of #2579.

It's not a problem of intense parallel activity. The symptom I've witnessed is sustained high CPU and increasing memory consumption until it runs out of memory and crashes.

danielbrunt57:
But the helper function has never been altered so I'm trying to understand what external change has caused the helper to now be in an infinite loop.  If you don't know that's fine as processing a copy of the list when the original list is going to be changed by the process is the correct way otherwise you usually end up shooting yourself in the foot!

dpwood:
> But the helper function has never been altered so I'm trying to understand what external change has caused the helper to now be in an infinite loop. If you don't know that's fine as processing a copy of the list when the original list is going to be changed by the process is the correct way otherwise you usually end up shooting yourself in the foot!

It was altered:

https://github.com/alandtse/alexa_media_player/commit/6f9a92de6fda7fd5361e15ae719c11092eaf7228#diff-4e8bc1554181c711fd75da33fbb2b0e7a8a1abd5096c723a1d3baa9f3c10543c

alandtse:
Went with #2580

Issue #2575, opened 2024-10-02
HAC TTS not working; Alexa device says "I am currently having problems to access your skill Simon Says at this moment"


nighto:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

When trying to TTS feature on Home Assistant Cloud, the TTS plays normally on the browser, however when using the Echo Dot device, it instead says "I am currently having problems to access your skill Simon Says at this moment" (or something similar, as the message was said in the language configured for the Alexa to speak, which in my case is Brazilian Portuguese.

Integration is properly working as I can otherwise control the device, check reminders etc.

**To Reproduce**

1. Go to `[your home assistant url]/config/cloud/account`
2. On the TTS card, click on Try
3. Choose browser to see it works on the browser
4. Choose Echo Dot and hear the error

**Expected behavior**

That the Echo Dot would play the audio generated by the TTS

**Screenshots**

(not applicable)

**System details**

- Home Assistant version: 2024.9.3
- alexa_media version (from `const.py` or HA startup log): (not sure how can I check)
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log): (not sure how can I check)
- Is Amazon [2FA/2SV enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account)  (y/n): Yes
- Amazon Domain: Amazon.nl

**Debug Logs (alexa_media & alexapy)**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

Followed FAQ instructions to enable log but I am not sure how to check, sorry

**Additional context**

nighto:
Got it: there was a final `/` missing on public url under Alexa Media Player integration. Sorry for the noise but hope this closed issue is helpful for folks having a similar issue on the future :)

PR #2588, opened 2024-10-03
fix: ensure public_url ends with trailing slash

danielbrunt57:
Fixes [issue #2575](https://github.com/alandtse/alexa_media_player/issues/2575)


nighto:
Thanks :)

PR #2589, opened 2024-10-04
fix: fix infinite loop and memory exhaustion

### LINKED ISSUE #2579: Version v4.13.3 is making HA fail

dpwood:
Iterate on a copy of the existing_serials rather than extend the list being iterated causing infinite iteration

Fixes #2579

### DIFF
--- a/custom_components/alexa_media/helpers.py
+++ b/custom_components/alexa_media/helpers.py
@@ -242,7 +242,7 @@ def _existing_serials(hass, login_obj) -> list:
             .get("devices", {})
             .get("media_player", {})
         )
-        for serial in existing_serials:
+        for serial in existing_serials[:]:
             device = device_data.get(serial, {})
             if "appDeviceList" in device and device["appDeviceList"]:
                 apps = [


danielbrunt57:
@dpwood Are you aware of @jleinenbach's PR to address #2579?  
I was not having an issue with the original code (or even 4.13.3 for that matter) so I'm wondering what conditions are present for an infinite loop to occur...

dpwood:
Aware of it but not tried it. It looked like an AI attempt at a solution rather than anything focused. A lot of changes that don't seem to address the actual problem. 

I was having the issue very reliably and the mistake seemed clear. This one line fix has resolved it for me.

Your question about conditions required for failure is a good one. I suspect number or types of Alexa devices is relevant? I'll do some more investigating.

dpwood:
On startup the existing_serials is [], after a few minutes the function re-runs and in my case I have 17 entries before entering the loop, which is the number of devices I have.  I then see it enter an infinite loop and the existing_serials array grows indefinitely.

I suspect the precondition is having a device with an appDeviceList which is somehow self-referencing? The new way of adding to the list is potentially recursive.

danielbrunt57:
The new code might be appending with a serial# vs without a serial# before? I suspect my environment (10 physical "included" Echo devices) might not have any appDeviceList entities. I wonder if it might be the Alexa App "device", which I've excluded (by not including it)...

dpwood:
I don't think the serial# behaviour has changed, only the way of building the list.

`existing_serials = existing_serials + apps`

will have been creating a new list each time, so wouldn't suffer this issue.

I've got Echo Dots, Echo Shows, Sonos, Fire TV sticks, and Alexa Mobile Voice iOS devices in the list. One of those is probably projecting an appDeviceList which has a serial that is then self-referencing.

I added logging to confirm that the issue is definitely within this loop. I can easily replicate it (happens every time). It is also resolved with this change.

alandtse:
Thanks!

jleinenbach:
I just want to say sorry that I was unable to fix it fast. I tested the version, but not long enough as it seems. I am still trying to improve the code, but I will test it much longer now!

PR #2634, opened 2024-10-19
fix: bump alexapy to 1.29.3

srescio:
tested in local dev instance that login loop is indeed fixed with the bump

addresses the following
https://gitlab.com/keatontaylor/alexapy/-/releases/v1.29.3

mapped in this repo as
https://github.com/alandtse/alexa_media_player/issues/2514
https://github.com/alandtse/alexa_media_player/issues/2515

### DIFF
--- a/custom_components/alexa_media/const.py
+++ b/custom_components/alexa_media/const.py
@@ -15,7 +15,7 @@ from homeassistant.const import (
     PERCENTAGE,
 )
 
-__version__ = "4.13.4"
+__version__ = "4.13.5"
 PROJECT_URL = "https://github.com/alandtse/alexa_media_player/"
 ISSUE_URL = f"{PROJECT_URL}issues"
 NOTIFY_URL = f"{PROJECT_URL}wiki/Configuration%3A-Notification-Component#use-the-notifyalexa_media-service"
--- a/custom_components/alexa_media/const.py
+++ b/custom_components/alexa_media/const.py
@@ -15,7 +15,7 @@ from homeassistant.const import (
     PERCENTAGE,
 )
 
-__version__ = "4.13.5"
+__version__ = "4.13.4"
 PROJECT_URL = "https://github.com/alandtse/alexa_media_player/"
 ISSUE_URL = f"{PROJECT_URL}issues"
 NOTIFY_URL = f"{PROJECT_URL}wiki/Configuration%3A-Notification-Component#use-the-notifyalexa_media-service"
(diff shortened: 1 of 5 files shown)


vhuynen:
Kudo @srescio,

I confirm that the alexapy **1.29.3** library fixes the login loop screen issue. To address this before the official release, you can upgrade the alexapy library in the file /homeassistant/custom_components/alexa_media/manifest.json:
`"requirements": ["alexapy==1.29.3", "packaging>=20.3", "wrapt>=1.14.0"]`
Restart Home Assistant and complete the OAuth flow to retrieve the refresh token used for API calls.

My settings HA :
Core : 2024.10.3
Supervisor : 2024.10.2
Operating System : 13.2

custom_components/alexa_media/manifest.json
```diff
@@ -8,6 +8,6 @@
   "iot_class": "cloud_polling",
   "issue_tracker": "https://github.com/alandtse/alexa_media_player/issues",
   "loggers": ["alexapy", "authcaptureproxy"],
-  "requirements": ["alexapy==1.29.2", "packaging>=20.3", "wrapt>=1.14.0"],
-  "version": "4.13.4"
+  "requirements": ["alexapy==1.29.3", "packaging>=20.3", "wrapt>=1.14.0"],
+  "version": "4.13.5"
```
alandtse: Do not touch the component version. That is handled by ci.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "alexa_media_player"
-version = "4.13.4"
+version = "4.13.5"
```
alandtse: Same as above.

alandtse (changes requested):

custom_components/alexa_media/const.py
```diff
@@ -15,7 +15,7 @@
     PERCENTAGE,
 )
 
-__version__ = "4.13.4"
+__version__ = "4.13.5"
```
srescio: just for confirmation @alandtse does it apply also to `const.py` here?

alandtse: Yes. The version is handled by ci. Any mismatch will break it.

srescio: i take its managed by [this action](https://github.com/alandtse/alexa_media_player/blob/dev/.github/workflows/semantic_release.yml#L79-L80)

Wizard99it:
@srescio Sei un grande 💪🏼 Grazie, fix  4.13.5 working until now!

PR #2648, opened 2024-10-26
test: add pytest to reproduce infinite loop bug

brianegge:
test: Add pytest to reproduce infinite loop bug. Move forward versions in order to test against homeassistant 'groups' object. Attempt to setup testing and coverage workflows.

Copied setup from https://github.com/bramstroker/homeassistant-powercalc


alandtse:
Title is way too long, but I'll shorten it. Thanks.

PR #2676, opened 2024-11-12
fix: public url missing

danielbrunt57:
This should fix issue [Upgraded to 4.13.8: Missing public url #2672](https://github.com/alandtse/alexa_media_player/issues/2672)

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -49,6 +49,7 @@ from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers import config_validation as cv, device_registry as dr
 from homeassistant.helpers.discovery import async_load_platform
 from homeassistant.helpers.dispatcher import async_dispatcher_send
+from homeassistant.helpers.network import NoURLAvailableError, get_url
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator, UpdateFailed
 from homeassistant.util import dt, slugify
 import voluptuous as vol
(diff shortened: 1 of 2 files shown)


alandtse:
Yah no idea. I'd assume the test handler has a circular import but no idea.

alandtse:
Wow, that's a lot of changes to the config flow. Does it have to be that much?

danielbrunt57:
Maybe I went overboard trying to address possible complications from yaml setup?
Personally, I'd rather deprecate legacy yaml setup with no automatic import and instruct users to remove it from their configuration, restart Home Assistant and use the UI to configure it instead.

In view of this, do not merge but close it instead and I'll do a new PR that reverts the changes in 4.13.8, unless you can just delete v4.13.8 and revert to 4.13.7 as the current version?

alandtse:
We can't just delete releases as it will confuse anyone who downloaded it.

alandtse:
Ok, so the issue I have here is this PR says it's fixing public urls but appears to do much more. Please make sure you do things atomicly so that we can easily revert things if need be. If you want to just do a revert of the last pr, you can submit a PR for that. Then we can do whatever you're doing to the import code. Again, the import code really shouldn't be used by anyone as we haven't really supported yaml for like 3 years.

danielbrunt57:
Okay, I'll do a PR with code from 4.13.7 to undo 4.13.8.

As for legacy yaml, I think it best to deprecate it and I'll not worry about trying to fix errors related to it.
I'll submit a new PR which initiates the warnings.  What version should I state when it will no longer work? 4.13.x? 4.14.0?

PR #2680, opened 2024-11-16
fix: revert missing public_url (#2670)

danielbrunt57:
This reverts `__init__.py` to v4.13.7


alandtse:
Git has a revert command. Is there a reason you're doing it manually?

danielbrunt57:
I did not know about a revert command.  I’m only 4 months into the world of Git!

alandtse:
Let's make it a real revert. https://ohshitgit.com/

danielbrunt57:
Well that was easy!  I had to google how to do it in UI and now that I'm aware of the button near the bottom of the PR, it's blatantly obvious but I've never scrolled down that far before or if I have, it never registered that I might need it someday!

PR #2681, opened 2024-11-16
fix: update strings.json

danielbrunt57:
- Removed: services: clear_history
- Changed: miscellaneous wording changes

### DIFF
--- a/custom_components/alexa_media/strings.json
+++ b/custom_components/alexa_media/strings.json
@@ -88,22 +88,5 @@
         }
       }
     }
-  },
-  "issues": {
-    "restart_required": {
-      "title": "Restart required",
-      "fix_flow": {
-        "step": {
-          "confirm_restart": {
-            "title": "Restart required",
-            "description": "Restart of Home Assistant is required to finish download/update of {name}, click submit to restart now."
-          }
-        }
-      }
-    },
-    "deprecated_yaml_configuration": {
-      "title": "YAML configuration is deprecated",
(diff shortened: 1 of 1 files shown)


alandtse:
This seems to be tied to another PR if you're introducing a warning `deprecated_yaml_configuration`.

danielbrunt57:
Yes, it contains the strings addition required for PR [#2685]( https://github.com/alandtse/alexa_media_player/pull/2685)

alandtse:
So changes should be atomic. So any string changes for that pr should be in the PR. You shouldn't require two PRs to get the effect.

danielbrunt57:
Alright,  I'll amend both PR's then...

danielbrunt57:
My thinking was that both PR's would be merged at the same time thus no harm.  Or the strings PR merged first then the deprecated PR following that, but I'll conform...

alandtse:
Also, after the other one goes through, please explain why you're updating the strings better.  Update doesn't explain anything. If we needed to review this 3 months later, it doesn't explain what has changed any more than the fact a file has been changed. It forces someone to actually read the changes to guess what was intended.

danielbrunt57:
If you prefer, I can cancel my PR for deprecate and let you merge strings.  Then I update my deprecate branch to your new dev and submit a new PR with new strings addition and config changes for deprecation...

danielbrunt57:
I'm going to close this one as it is basically just cosmetic + remove a deprecated section.
I think it's more important to move forward on the yaml deprecation and the 4.13.8 reversion.

PR #2685, opened 2024-11-17
fix!: deprecate YAML configuration

danielbrunt57:
YAML configuration is deprecated. Users must manually remove `alexa_media` from `configuration.yaml` file and add the integration via UI to avoid warning messages and future failure.

BREAKING CHANGE: Deprecate YAML configuration

### DIFF
--- a/custom_components/alexa_media/strings.json
+++ b/custom_components/alexa_media/strings.json
@@ -104,17 +104,6 @@
     }
   },
   "issues": {
-    "restart_required": {
-      "title": "Restart required",
-      "fix_flow": {
-        "step": {
-          "confirm_restart": {
-            "title": "Restart required",
-            "description": "Restart of Home Assistant is required to finish download/update of {name}, click submit to restart now."
-          }
-        }
-      }
-    },
     "deprecated_yaml_configuration": {
       "title": "YAML configuration is deprecated",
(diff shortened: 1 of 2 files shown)


danielbrunt57:
Requires PR #2681

alandtse:
To be clear, the only changes to strings should be related to the deprecation.

custom_components/alexa_media/strings.json
```diff
@@ -102,5 +88,22 @@
...
+          "confirm_restart": {
+            "title": "Restart required",
+            "description": "Restart of Home Assistant is required to finish download/update of {name}, click submit to restart now."
+          }
+        }
+      }
+    },
+    "deprecated_yaml_configuration": {
+      "title": "YAML configuration is deprecated",
+      "description": "YAML configuration of Alexa Media Player is deprecated\nand will be removed in version 4.14.0.\nThere will be no automatic import of this.\nPlease remove it from your configuration, restart Home Assistant and use the UI to configure it instead.\nSettings > Devices & services > Integrations > ADD INTEGRATION"
```
alandtse, line 109: Everything else in the string.json is inappropriate for this commit.

custom_components/alexa_media/strings.json
```diff
@@ -102,5 +102,22 @@
         }
       }
     }
+  },
+  "issues": {
+    "restart_required": {
```
alandtse: Where is this used?

danielbrunt57: I copied code from HACS when they deprecated YAML but I realize now that my other edits don't use it (now that you've pointed it out) and have removed that section.

alandtse (approved):

Issue #2709, opened 2024-12-02
Duplicated entities when using Home-Assistant-Matter-Hub custom Add-on


digiolin:
**Describe the bug/issue**

I started to use the custom add-on [Home-Assistant-Matter-Hub](https://github.com/t0bst4r/home-assistant-matter-hub/) which creates a Matter Hub whereby HA devices can be controlled via Alexa without using the [Alexa via Home Assistant Cloud](https://www.home-assistant.io/cloud/alexa/) via Nabu Casa or the manually created [Amazon Alexa Smart Home Skill](https://www.home-assistant.io/integrations/alexa.smart_home/) (very slick!)

The Alexa Media Player application has logic already to handle devices added via the official [Alexa Integration](https://www.home-assistant.io/integrations/alexa.smart_home/) when using `Include devices connected via Echo`.

I duplicated [this issue](https://github.com/t0bst4r/home-assistant-matter-hub/issues/245) here so it can be a combined effort to find a good solution!

Here an example of a debug logs from a switch device:

```
{
    "applianceId": "AAA_SonarCloudService_<void>",
    "endpointTypeId": "",
    "driverIdentity": {
        "namespace": "AAA",
        "identifier": "SonarCloudService"
    },
    "manufacturerName": "TestVendor",
    "friendlyDescription": "TestVendor smart device",
    "modelName": "",
    "deviceType": "",
    "version": "0",
    "friendlyName": "Camara",
    "friendlyNameModifiedAt": 1733084691197,
    "capabilitiesModifiedAt": 1733084691197,
    "ipAddress": "",
    "port": "",
    "entityId": "<void>",
    "applianceNetworkState": {
        "reachability": "REACHABLE",
        "lastSeenAt": 1733130114179,
        "createdAt": 1733084691197,
        "lastSeenDiscoverySessionId": {
            "value": "<void>"
        }
    },
    "tags": {
        "tagNameToValueSetMap": {}
    },
    "additionalApplianceDetails": {
        "additionalApplianceDetails": {}
    },
    "firmwareVersion": "0",
    "actions": [],
    "capabilities": [
        {
            "capabilityType": "AVSInterfaceCapability",
            "type": "AlexaInterface",
            "version": "3",
            "properties": {
                "supported": [
                    {
                        "name": "powerState"
                    }
                ],
                "proactivelyReported": true,
                "retrievable": true,
                "readOnly": false
            },
            "interfaceName": "Alexa.PowerController"
        },
        {
[...]

PR #2730, opened 2024-12-06
fix: home-assistant-matter-hub duplicated entities

danielbrunt57:
Force "TestVendor" in the `is_local` check to return `False` since the entities discovered by Echo devices via Matter discovery are duplicates of already existing HA entities.

Fixes issue #2709


danielbrunt57:
If @t0bst4r fixes his `manufacturerName` so that it's `t0bst4r` rather than `TestVendor`, this will have to be updated again in `alexa_entity.py`.

danielbrunt57:
FYI, this code change with Home-Assistant-Matter-Hub 3.0.0-alpha.52 is working as intended and preventing duplication of entities.

PR #2732, opened 2024-12-08
Clear error message for music streaming

sayam93:
This clearly conveys to the users that the inability to stream music is not a problem with the integration or home assistant but instead something that Amazon doesn't allow.

### DIFF
--- a/custom_components/alexa_media/media_player.py
+++ b/custom_components/alexa_media/media_player.py
@@ -1375,7 +1375,7 @@ class AlexaClient(MediaPlayerDevice, AlexaMedia):
             )
 
     @_catch_login_errors
-    async def async_play_tts_cloud_say(self, public_url, media_id, **kwargs):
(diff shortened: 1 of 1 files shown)


alandtse (commented):
Thanks. I appreciate what you're trying to fix, but the message was also created to address another use case. I believe where people type in TTS in the media player card and it didn't work. We should not completely remove the old message.

If possible, a solution should detect each case and provide the right message. That may not be possible. Can you double check and then address both the old and new message please?

sayam93:
> Thanks. I appreciate what you're trying to fix, but the message was also created to address another use case. I believe where people type in TTS in the media player card and it didn't work. We should not completely remove the old message.
> 
> If possible, a solution should detect each case and provide the right message. That may not be possible. Can you double check and then address both the old and new message please?

I have made a second commit by changing the logic a bit to handle this. If media type is music and no public url is provided, it will play the old message. However, if the public url is provided when the media type is music, it will tell the user that they cannot stream music.

PR #2742, opened 2024-12-10
fix: TTS Speak Action in Home Assistant

sayam93:
Fixes the broken TTS action in Home Assistant that was broken in 5.0.3

TTS is defined as media_type music by the TTS component in Home Assistant which resulted in breaking the tts speak action. However, the async_speak function in the TTS component in HA always sets ATTR_MEDIA_ANNOUNCE to true in order to differentiate it from music (https://github.com/home-assistant/core/blob/b46392041f36cc932d0a12eb43af20ecfb7f25db/homeassistant/components/tts/__init__.py#L475) which is now properly implemented.

### DIFF
--- a/custom_components/alexa_media/const.py
+++ b/custom_components/alexa_media/const.py
@@ -111,13 +111,12 @@ ATTR_MESSAGE = "message"
 ATTR_EMAIL = "email"
 ATTR_NUM_ENTRIES = "entries"
 STREAMING_ERROR_MESSAGE = (
-    "Sorry folks! Amazon doesn't allow streaming music like this. "
-    "Please take it up with them!"
+    "Sorry, direct music streaming isn't supported. "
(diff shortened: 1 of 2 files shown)


alandtse:
Is the issue we can't differentiate TTS vs music? We don't want to lose the warning message to set public_url for those doing TTS.  Do we need both warnings in the text?

sayam93:
> Is the issue we can't differentiate TTS vs music? We don't want to lose the warning message to set public_url for those doing TTS. Do we need both warnings in the text?

TTS is defined as media_type music by the TTS component in Home Assistant which resulted in breaking the tts speak action. However, the async_speak function in the TTS component in HA always sets ATTR_MEDIA_ANNOUNCE to true in order to differentiate it from music (https://github.com/home-assistant/core/blob/b46392041f36cc932d0a12eb43af20ecfb7f25db/homeassistant/components/tts/__init__.py#L475) which is now properly implemented.

This does not remove or alter the warning for public url.

PR #2746, opened 2024-12-12
feat: readd support for Amazon devices

alams154:
Reintroduces Amazon manufactured devices to be exposed by the integration.  This is especially useful for Amazon Smart Plugs which cannot be added to Home Assistant via any other means.

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -89,6 +89,7 @@ def is_local(appliance: dict[str, Any]) -> bool:
     if (
         appliance.get("manufacturerName") == "Ledvance"
         or appliance.get("manufacturerName") == "Sengled"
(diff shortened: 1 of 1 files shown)


alandtse:
Thanks. Can we simplify the test so it's 
```py
appliance.get("manufacturerName") in ["x","y","z"]

danielbrunt57:
> Reintroduces Amazon manufactured devices to be exposed by the integration.

When was it removed??

Your modification broke the code for AIAQM sensors!!

danielbrunt57:
Aren't Amazon Smart Plugs similar to Air Quality Monitors in that they exist in Amazon Alexa but are not connected via an Echo device?  I would think it would make more sense to modify the `parse_alexa_entities` to process them like it does the AIAQM sensors rather than fudging the `is_local` test since they aren't 'local' in the original design concept of `alexa_entity.py`...

PR #2773, opened 2024-12-26
feat: Add restore volume service

danielbrunt57:
This adds a new service: **_Alexa Media Player: Restore Previous Volume_** (`alexa_media.restore_volume`)


alandtse:
Please update the wiki.

dpgh947:
I didn't think this warranted opening an issue (yet), but wanted to explain a possible problem I found with the new volume save/restore function. I use node red for all my automation stuff, and already had a notify flow that allows for loud TTS if requested. It uses a current state node to save the current volume level, then sets the loud volume, issues the notify, then sets the volume back. In a tinkering mood I thought I would update this to use the new restore volume action. I tested that a set volume call populated the previous volume attribute, and then calling the restore action did indeed restore it. But when put in to my flow, I found that the previous volume attribute was not being updated. On a hunch I thought this might be a timing issue caused by the volume set being followed immediately by the notify, and put a delay in between. 1 second made no difference, but 2 seconds made it work. So there is a timing issue somewhere in the mix - a set volume call from node red followed immediately by a TTS notify call prevents the previous volume attribute update somehow, but a 2 second delay in between makes it work. Any thoughts?

dpgh947:
I'm calling media_player.volume_set

danielbrunt57:
> a set volume call from **_node red_** followed immediately by a TTS notify call prevents the previous volume attribute update somehow, but a 2 second delay in between makes it work.

danielbrunt57:
I'll gladly revert the changes if people are not happy with it and someone else can tackle it.

alandtse:
@danielbrunt57 thanks for all your help. Do you need a break? You're sounding like burn out imho. I'm only seeing some of the chatter, but remember it's ok to step away.

danielbrunt57:
Yes I do!  My townhouse sold last week and I now have until the end of February to pack up and vacate!  I've already de-automated most of my motion sensors, lights, switches and door locks.  It sure is a pain having to remember to turn on a light when entering a room and it's even tougher to remember to turn it off afterwards!!

dpgh947:
> You are contradicting yourself.

I'm really not. I'm calling that action from node red. I could put that in italics if it helps.

image

PR #2791, opened 2025-01-06
fix: Optimize OptionsFlow (#2748)

z-master42:
Make #2748 compatible with Home Assistant version < 2024.12 and be more consistent in the implementation of https://developers.home-assistant.io/blog/2024/11/12/options-flow


alandtse:
Is this an actual fix of bad behavior or is it just an improvement? It determines whether this will cause a release. If it's just an improvement, I'd change to `refactor` or `perf` depending on the expected impact.

danielbrunt57:
> Make #2748 compatible with Home Assistant version < 2024.12

It looks more like a `fix` to me than an improvement if #2748 fails to work on HA < 2024.12...

PR #2826, opened 2025-02-20
refactor: require OTPSECRET 52-character key

danielbrunt57:
BREAKING CHANGE: Force Amazon 2-step Verification via Authenticator App secret (Generate OTP using an application). 
If not previously configured this way, this will required removing and re-installing Alexa Media Player to configure it with the now mandatory 52-character Authenticator App secret so that AMP can generate its own TOTP to authenticate your Amazon account credentials.  You will also have to enable your Amazon account to require an authenticator app generated OTP to log in. See the Wiki for further details.

### DIFF
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -479,7 +479,7 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
                     password=self.config[CONF_PASSWORD],
                     outputpath=self.hass.config.path,
                     debug=self.config[CONF_DEBUG],
-                    otp_secret=self.config[CONF_OTPSECRET]
+                    otp_secret=self.config[CONF_OTPSECRET],
                     uuid=uuid,
                     oauth_login=True,
                 )
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -955,7 +955,9 @@ class OptionsFlowHandler(config_entries.OptionsFlow):
                     CONF_SECURITYCODE
                 ]
             if CONF_OTPSECRET in self.config_entry.data:
-                user_input[CONF_OTPSECRET] = self.config_entry.data[CONF_OTPSECRET].replace(" ","")
+                user_input[CONF_OTPSECRET] = self.config_entry.data[
+                    CONF_OTPSECRET
+                ].replace(" ", "")
             if CONF_OAUTH in self.config_entry.data:
                 user_input[CONF_OAUTH] = self.config_entry.data[CONF_OAUTH]
             """Ensure public_url ends with trailing slash"""
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -137,7 +137,7 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
                 (vol.Required(CONF_HASS_URL), str),
                 (vol.Required(CONF_EMAIL), str),
                 (vol.Required(CONF_PASSWORD), str),
-                (vol.Optional(CONF_OTPSECRET), str),
+                (vol.Required(CONF_OTPSECRET), str),
                 (vol.Optional(CONF_SECURITYCODE), str),
                 (vol.Optional(CONF_PUBLIC_URL), str),
                 (vol.Optional(CONF_INCLUDE_DEVICES, default=""), str),
@@ -197,7 +197,7 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
                     str,
                 ),
                 (
-                    vol.Optional(
+                    vol.Required(
                         CONF_OTPSECRET, default=self.config.get(CONF_OTPSECRET, "")
                     ),
                     str,
@@ -355,7 +355,7 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
         ):
             otp: str = self.login.get_totp_token()
             if otp:
-                _LOGGER.debug("Generating OTP from %s", otp)
+                _LOGGER.debug("Generated OTP token: %s", otp)
                 return self.async_show_form(
                     step_id="totp_register",
                     data_schema=vol.Schema(self.totp_register),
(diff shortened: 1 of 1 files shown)


jleinenbach:
I requested ChatGPT o3-mini-high to review the changes. @danielbrunt57 – could you please verify the "Schema Consistency for CONF_OTPSECRET" section?

---

### 1. OptionsFlowHandler Initialization for HA 2025 Compatibility

**Issue:**  
Home Assistant 2025 expects the configuration entry to be available in the options flow. In older versions, some developers used conditional logic (e.g. only setting `self.config_entry` when HA version is below a certain threshold). However, not setting it in newer versions leads to errors when accessing `self.config_entry.data`.

**Proposed Improvement:**  
To maintain backward compatibility while satisfying the new API requirements, you can conditionally assign the config entry. For example:

```python
from awesomeversion import AwesomeVersion
from homeassistant.const import __version__ as HAVERSION

class OptionsFlowHandler(config_entries.OptionsFlow):
    def __init__(self, config_entry: config_entries.ConfigEntry) -> None:
        """Initialize options flow with backward compatibility."""
        # For older HA versions, you might have conditioned this assignment.
        # However, in HA 2025 and later, the config_entry must always be available.
        self.config_entry = config_entry
        self.config = OrderedDict()
```

**Rationale:**  
This approach ensures that `self.config_entry` is always set regardless of the Home Assistant version. It preserves compatibility with older versions (which might have expected an explicit assignment) while meeting the requirements of HA 2025.

---

### 2. Logging Formatting

**Issue:**  
A log call such as:
```python
_LOGGER.debug("Generated OTP token", otp)
```
does not include a formatting placeholder (e.g. `%s`). Without a placeholder, the OTP token will not be inserted into the log message, and you might receive formatting warnings.

**Proposed Improvement:**  
Change the log call to:
```python
_LOGGER.debug("Generated OTP token: %s", otp)
```

**Rationale:**  
This ensures that the OTP token is correctly inserted into the log message, improving readability and debugging.

---

### 3. Schema Consistency for CONF_OTPSECRET

**Issue:**  
There is an inconsistency in the schema definition for `CONF_OTPSECRET`. In some parts, it is declared as required (using `vol.Required`), while in others it is treated as optional (using `vol.Optional`). This can lead to validation errors or force users to input a value unnecessarily.

**Proposed Improvement:**  
Decide clearly whether `CONF_OTPSECRET` should be required or optional. For example, if it is optional, update the schema consistently:

```python
(vol.Optional(CONF_OTPSECRET, default=""), str)
```

**Rationale:**  
A consistent declaration prevents unexpected validation errors and improves user experience by not forcing unnecessary input.

---

### 4. Safe URL Substring Extraction

**Issue:**  
The code uses `slugify(login.url[7:])` to generate part of a notification ID, assuming that `login.url` is at least 7 characters long. If `login.url` is shorter, this will raise an IndexError.

**Proposed Improvement:**  
Perform a length check before slicing the URL. For example:

```python
url_substr = login.url[7:] if len(login.url) > 7 else login.url
notification_id = f"alexa_media_{slugify(email)}{slugify(url_substr)}"
```

**Rationale:**  
This change ensures that the substring extraction is safe even if `login.url` is unexpectedly short, thereby avoiding potential runtime errors.

---

### Additional Considerations

- **PEP 8 Formatting & Docstrings:**  
  Ensure that the code follows PEP 8 guidelines (proper indentation, line length, etc.) and includes complete docstrings for all methods and classes.

- **Type Annotations:**  
  Review and, if necessary, add consistent type annotations for function parameters and return types to improve code clarity and maintainability.

- **2FA Logic and `automatic_steps`:**  
  The current handling of two-factor authentication (2FA) via `automatic_steps` is complex and error-prone. Consider refactoring this logic for clarity, such as separating automatic retries from user-driven input.

- **_update_ord_dict Method:**  
  The current implementation of `_update_ord_dict` is not optimal. You might simplify it by iterating over the keys of the old dictionary and replacing them with the new values if they exist, or consider using dictionary comprehensions if order is not critical.

- **Usage of `reduce`:**  
  Replace the use of `reduce` for list-to-string conversion with Python's `join` method. For example, instead of:
  ```python
  reduce(lambda x, y: f"{x},{y}", user_input[CONF_INCLUDE_DEVICES])
  ```
  use:
  ```python
  ",".join(user_input[CONF_INCLUDE_DEVICES])
  ```

---

### Summary

- **OptionsFlowHandler:** Always set `self.config_entry` to ensure compatibility with both older versions and HA 2025.
- **Logging:** Use proper format placeholders (e.g. `%s`) to include dynamic values in log messages.
- **Schema Consistency:** Define `CONF_OTPSECRET` uniformly (either as required or optional) across the code.
- **URL Handling:** Add length checks when slicing URLs to avoid runtime errors.
- **Other improvements:** Ensure PEP 8 compliance, use clear docstrings, consistent type annotations, simplify list-to-string conversion with `join`, and consider refactoring complex 2FA logic.

These changes are aimed at making the code robust, maintainable, and future-proof for upcoming Home Assistant releases.

danielbrunt57:
Hi @jleinenbach I had a quick look at your code while sitting in my car plugged in and charging in my garage (for 4 more days) at 12:23 am in an otherwise vacanct house with all my worldly possessions packed up and sitting in either storage lockers or on two U-Haul trailers parked in New Westminster!!!
I'll leave it up to you or Alan to adjust it, or abandon it.  I've just see too many failures when not using OTP app that go away with using an app secret...

danielbrunt57:
> * **2FA Logic and `automatic_steps`:**
>   The current handling of two-factor authentication (2FA) via `automatic_steps` is complex and error-prone. Consider refactoring this logic for clarity, such as separating automatic retries from user-driven input.

This is intriguing!  Maybe there's a way to fix that stupid sequence that broke when Amazon restructured the login windows separating email and password.  Even better to replace that with the proxy method that Alan suggested months ago...

alandtse:
Unless we have strong indications this is the solution, I'm inclined to hold off. Right now you're seeing a certain subset of people with problems but not those who are working fine. A universal breaking change may break the people working fine using other otp mechanisms.

That said I'm fine not supporting people who don't use the built in authenticator. We already make statements about needing 2fa specifically to avoid Amazon's step up authentication with JavaScript.

danielbrunt57:
> What's the thought behind making this mandatory? This will have substantial impact across the user base.

This is one of many ongoing reasons why...
https://github.com/alandtse/alexa_media_player/issues/2848#issuecomment-2707507079

PR #2832, opened 2025-02-24
perf: select the same URL for NotifyNowPlayingUpdated as for get_state()

nao-pon:
Selecting the same URL prevents unnecessary HTTP requests.

### DIFF
--- a/custom_components/alexa_media/media_player.py
+++ b/custom_components/alexa_media/media_player.py
@@ -750,7 +750,7 @@ class AlexaClient(MediaPlayerDevice, AlexaMedia):
                 )
                 if self._media_image_url is None:
                     self._media_image_url = (
-                        self._session.get("mainArt", {}).get("largeUrl")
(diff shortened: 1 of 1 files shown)


alandtse:
I'm not sure how this is a performance tweak. We're just providing a different URL that had already been retrieved. Is there a risk the new URL won't exist?

nao-pon:
Below are the HTTP requests when trying both largeURL and fullURL.



In this example, largeURL is the image h256.w256. By contrast, get_state() is h600.w600. First the h256.w256 image is loaded, and then h600.w600 is loaded. The image flickers when it is loaded.



In this example, both fullURL and get_state() are h600.w600, so no duplicate images are loaded and no flickering occurs.

nao-pon:
In the first place, I don't think there is any need to call get_stare(), but I haven't been able to trace clearly where it is being called from.

nao-pon:
I've looked into the source code and found that I was inadvertently calling get_state(). I'm closing this PR and submitting a new one.

PR #2776, opened 2024-12-28
feat: Add restore volume service (#2773)

alandtse:

PR #2875, opened 2025-03-25
feat: add get_history_records service call

Joe91:
added a method to get some of the history-records. Need this for one of mine automations, maybe someone else can use this too.
I don't know your requirements or target structure, so just give me feedback.
If this is not a desired behavior I'm also fine with this.

Thanks a lot for this amazing component!
Joe


custom_components/alexa_media/services.py
```diff
@@ -190,3 +208,66 @@ async def restore_volume(self, call) -> bool:
...
+            _LOGGER.error("Entity %s not found or not part of %s", entity_id, DOMAIN)
+            return False
+        target_serial_number = entity_entry.unique_id
+
+        history_data_total = []
+        for email, account_dict in self.hass.data[DATA_ALEXAMEDIA]["accounts"].items():
+            login_obj = account_dict["login_obj"]
+            try:
+                # Get the history records. Input: Time from, Time to,
+                history_data = await AlexaAPI.get_customer_history_records(
```
alandtse, line 235: Just to confirm, alexapy already has this? Or do we need an alexapy bump?

Joe91: Yeah, it's already there.
Found it when I thought I had to write it on my own. 
Quite a powerful feature in my opinion...

danielbrunt57: Yes, `alexaapi.py` already has this and the `update_last_called` service uses it via:
```
    async def get_last_device_serial(
        login: AlexaLogin, items: int = 10
    ) -> Optional[dict[str, Any]]:
        """Identify the last device's serial number and last summary.

        This will search the [last items] activity records and find the latest
        entry where Echo successfully responded.
        """
        response = await AlexaAPI.get_customer_history_records(
            login, max_record_size=items
        )
```

Joe91: ah okay. But it's not already used in the media_player. Or is it?

danielbrunt57: I can see the possible merit in doing what you're doing.  I'm just not quite seeing how or where you're going to use the returned data.
There's a fundamental underlying current issue with the HTTP "GET" history records though and that is when using the original HTTP path `customer_history_records`, it almost always has to retry the GET (up to 5 times) and often fails with "Too Many Requests".  That HTTP access point is what the now defunct web interface to Alexa used to use.  Apollon77's `alexa-remote` (javascript) now uses `customer_history_records_v2` and can almost always GET records repeatedly without the dreaded "429" error.  I tried rewriting `alexapy` to use it but Apollon surrounds it with a different authentication scheme that I've not been able to unravel yet and convert to Python and I always get "403 Unauthorized".

Joe91: That returned data is stored in the new attribute "history_records".
I can then access this with templates or in the Python-Scripts.

I have one device that is only controllable with Alexa and this way I can evaluate the state of the device and control it depending on the state. 

What is your suggestion based on that issue? 
Since I'm already happy with this state for my usecase I'm totally fine to not merge that at all or to wait for your rework first.

Joe91:
stupid question: if this shall become a merge, what am I missing to fix this validation? Is there any mechanism for automatic translation into the different languages or what is required to fix this check?

custom_components/alexa_media/services.py
danielbrunt57: You could test modifying the existing service call to see if it meets your needs and has no downside.

Joe91: What do you mean with the existing service-call? I can't see any other option to access the history-records then using the api-call provieded.
Do you mean to improt the alexa-remote?

alandtse:
En.json should be populated I believe. Other languages are handled by lokalise.

danielbrunt57:
Correct.  See my [PR #2776](https://github.com/alandtse/alexa_media_player/pull/2776) from December where I had the same issue.

custom_components/alexa_media/services.py
danielbrunt57: > ah okay. But it's not already used in the media_player. Or is it?

> What do you mean with the existing service-call? I can't see any other option to access the history-records then using the api-call provided.

AMP calls `update_last_called` when AMP initializes at startup and then every time it polls Amazon, unless automatic polling is disabled by the user. That function calls `get_last_device_serial` which calls `get_customer_history_records`.
Your enhancement could be added inside `get_last_device_serial` to automatically update the new fields when polling occurs or the service `update_last_called` is manually called.  Having a new clearly defined service that updates the history records might be clearer for the user when making service calls manually but, increasing or duplicating automatic calls to `get_customer_history_records` will only add to its current "429 Too Many Requests" errors so we need to bear that in mind.

> Do you mean to import the alexa-remote?

No.  alexa-remote is javascript code so that's not possible.

Joe91: Ah, so you recommend to change this in alexapy module (to sore and provide it at the same moment) and afterwards acces it with alexa_media_player at this exact situation?
Edit: so it might already be enough to insert some more suff there and use this later up in this module:
https://gitlab.com/keatontaylor/alexapy/-/blob/dev/alexapy/alexaapi.py?ref_type=heads#L1491

Is that how you would solve this?

danielbrunt57:
After further fleshing the scenarios out in my head, we could update the description for the `update_last_called` to indicate it also updates history records when it's called thus making the new service call unnecessary as the existing service call is already retrieving customer history records.

danielbrunt57:
I would modify `get_last_device_serial` to also update the history records at the same time it's updating last_called since it's already retrieved the last 10 records.

Joe91:
I'm completely fine with this proposal :-).
For my automation I need especally the response of the last (better the last few) records. Since there is a lot more in these entries I would recommend to filter them and only use the entries that might be useful.

If they are already available (because of the exact same call being performed cyclic much deeper) the proposal of mine is stupid and will only cause additional traffic. I just didn't see / know that when I started.
Will you or someone else have a look at this? I'm also fine with if you say that this is not the direction you want this module to be, then I'll just keept this as a solution for myself.

Anyway this module is really amazing. Thanks for maintaining and supporting this!

alandtse:
As this is a service call, I'm ok with just adding the service. Most users will not be impacted   An optimization to auto populate can come later.

alandtse:
Thanks. I'll probably do a release this weekend. This should give a few days for translators to add the keys.

Joe91:
Great, thanks. 
I did a lot of testing and even though this works really great for me and the automation, I discovered two issues, that either need good documentation or a fix to make this usable for the default user.

- the async call to update the attribute will take a short period of time, so you have to consider this in the script. Changing this to await crashed the script since suddenly the call seems be to None, even if it's still working. I have no idea what's happening there. Is there a better way to update the attribute? 

- since the data doesn't seem to arrive in the media_player class in the internal attribute (don't know why or how to fix this) the history will disappear after a while again, haven't found out when exactly yet. I think it might be related to the first issue, that this is not a good way to set the attribute?

Really sorry for discovering these two issues after this was merged. Maybe someone can give me some hint in how to fix them or do it correctly?  I will provide a fix as soon as possible, but we can also revert this for now.

alandtse:
The wiki is used for documentation. Please document their. Attributes are refreshed based on events. That refresh could be clearing the data.

custom_components/alexa_media/services.py
danielbrunt57: No, it should be done in the AMP function that calls alexapy. alexapy can't update AMP config entry components. It simply returns x number of history records and it's up to the calling function to process the data it retrieved.  In AMP, function 1 calls function 2 which calls alexapy so it should be handled by either function 1 or 2 and from memory, I think function 2 is where I would add handling the data processing for the new feature.

danielbrunt57:
@alandtse Have you not read any of my comments?  I think your merging this as is was premature as I believe no consideration has been given to the already existing problem of 429 Too Many Requests when trying to update last called. If users try to also use this, the 429's are most likely going to increase.

danielbrunt57:
@alandtse So just refactor it down the road then. Okay. Agreed, it's not likely to get widespread use initially. I still want to dig deep into converting to V2 history records as that will impact everyone greatly. It'll be a challenge but one I'd like to conquer. Might take 6 months or longer. I'm still transitioning my life and currently have 205 sq ft of storage almost filled with my lifelong "shit" (as my ex calls it!) that I need to deal with before I buy and move into a trailer "somewhere" in BC. Once all that is done, I'll have time to "play" (as my ex calls it!) on my computer!

pedolsky:
@Joe91 , could you please give an example of using this new action?

Joe91:
here is a little example out of one of my python-scripts that uses this:

```
# Send voice command to Alexa
hass.services.call('media_player', 'play_media', {
    'entity_id': 'media_player.echo_dot',
    'media_content_id': media_content_id,
    'media_content_type': 'custom'
})

# get response
response = ""
timestamp = 0
for i in range(25): 
    time.sleep(1)
    # get the history records
    hass.services.call('alexa_media', 'get_history_records', {
        'entity_id': 'media_player.echo_dot'
    })
    state = hass.states.get('media_player.echo_dot')
    attributes = state.attributes
    response = ""
    
    for record in attributes['history_records']:
        if record['summary'] in media_content_id:
            response = record['response']
            timestamp = int(record['timestamp']/1000)
            break
    if response != "" and timestamp > now:
        break

```
With this I can extract the response and evalute it afterwards.

Issue #2868, opened 2025-03-17
Setting Volume error


DerGary:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

When setting Volume Home assistant stops the Script with an error although IT Changes the volume

**To Reproduce**

Call Action media_player.volume_set within a Script.

**Expected behavior**

The Script Changes Volume and Carries on without errors

**Screenshots**

![Image](https://github.com/user-attachments/assets/e404b358-2975-497b-bb00-3a8beb57f53b)

**System details**

- Home Assistant version: latest
- alexa_media version (from `const.py` or HA startup log): latest
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log):
- Is Amazon [2FA/2SV enabled](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account)  (y/n):
- Amazon Domain:

**Debug Logs (alexa_media & alexapy)**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

Logger: homeassistant.components.script.alexa_wecker_langsam_lauter
Quelle: helpers/script.py:524
Integration: Skript (Dokumentation, Probleme)
Erstmals aufgetreten: 07:06:17 (5 Vorkommnisse)
Zuletzt protokolliert: 07:26:34

Alexa Wecker Langsam Lauter: Sequence action at step 1: Repeat at step 4: Error executing script. Unexpected error for call_service at pos 1: 'volume'
Alexa Wecker Langsam Lauter: Sequence action at step 1: Error executing script. Unexpected error for repeat at pos 4: 'volume'
Alexa Wecker Langsam Lauter: Error executing script. Unexpected error for sequence at pos 1: 'volume'
Alexa Wecker Langsam Lauter: Sequence action at step 1: Error executing script. Unexpected error for call_service at pos 1: 'volume'
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 524, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 1009, in _async_step_call_service
    response_data = await self._async_run_long_action(
[...]

Issue #2889, opened 2025-04-07
Media player shuffle set stop working on echo media player


xitters:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

Since today I have noticed that Media Player Shuffle Set stopped working for an Echo Dot, the automation worked previously, without errors, but now something changed in Amazon Alexa that does not accept the Shuffle Set command?

2025-04-07 06:19:02.219 ERROR (MainThread) [homeassistant.components.automation.despertador] Despertador: Repeat at step 4: Error executing script. Error for call_service at pos 6: Entity media_player.echo_dot_de_juan does not support action media_player.shuffle_set
2025-04-07 06:19:02.219 ERROR (MainThread) [homeassistant.components.automation.despertador] Despertador: Error executing script. Error for repeat at pos 4: Entity media_player.echo_dot_de_juan does not support action media_player.shuffle_set
2025-04-07 06:19:02.220 ERROR (MainThread) [homeassistant.components.automation.despertador] Error while executing automation automation.despertador: Entity media_player.echo_dot_de_juan does not support action media_player.shuffle_set

automation with Media player shuffle set dont works now

**Expected behavior**

change something in the api of amazon alexa?

**Screenshots**

none

**System details**

- Home Assistant version:
- alexa_media version 5.6.0
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log):
- Is Amazon [2SV via Authenticator App](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (not 2SV OTP via SMS) enabled? Y
- Amazon Domain:

xitters:
It does not work with the last update 5.7.1

Error while executing automation automation.despertador: Entity media_player.echo_dot_de_juan does not support action media_player.shuffle_set

PR #2891, opened 2025-04-09
fix: fixed volume, supported_features issue in Version 5.5-6

nao-pon:
This may resolve issues #2868, #2889.

I haven't had the reported issues yet, so this needs validation from someone who does.

### DIFF
--- a/custom_components/alexa_media/media_player.py
+++ b/custom_components/alexa_media/media_player.py
@@ -28,7 +28,6 @@ from homeassistant.components.media_player.const import (
     MediaType,
     RepeatMode,
 )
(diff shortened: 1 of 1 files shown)


danielbrunt57:
Looks promising to me but I don't have the issue with only 2 active echos now. We'll have to see how it fares with others who have the issue.

JagerSprinkles:
I've been testing this for almost a week and have not had any issues, it's working perfectly for me. Not sure if anyone else with the listed issues had a chance to test this branch but I can confirm it fixed the #2868 issue for me.

Issue #2893, opened 2025-04-10
Reconfiguration Prompts 0 Devices after Media Player Failure


TonyMontuna:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

I was getting an error from my automations which trigger a voice alert through my alexa media players. I assumed there was something going on with the integration or something, so I deleted the configuration, re-logged in, created a new app authentication code (copying the 25 character code for 2fa app), pasted that into the configuration, hit submit, copied the code from the setup wizard, pasted it to the amazon two step website, confirmed it, then went back to hass, hit submit, logged in, authenticated with the 2fa code, and submit again. All was successful, except now, I have 0 entities. I used to have 17 alexa media player devices with over 100 entities. 

It does not load. 

To reproduce, at least in my case, delete the configuration from settings>integrations>alexa media player> delete [your@email] and then hit add new, log in, and hit submit. Nothing will appear. 

**Expected behavior**

create the entities associated with my alexa account 

**Screenshots**

**System details**

- Home Assistant version: 
- alexa_media version (from `const.py` or HA startup log): 5.6.0
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log): 412
- Is Amazon [2SV via Authenticator App (https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (not 2SV OTP via SMS) enabled?  (y/n): yes
- Amazon Domain: amazon.com

**Debug Logs (alexa_media & alexapy)**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

[home-assistant_alexa_media_2025-04-10T21-11-06.491Z.log](https://github.com/user-attachments/files/19694512/home-assistant_alexa_media_2025-04-10T21-11-06.491Z.log)

**Additional context**

This error originated from a custom integration.

Logger: custom_components.alexa_media
Source: helpers/update_coordinator.py:412
integration: Alexa Media Player (documentation, issues)
First occurred: 4:56:16 PM (2 occurrences)
Last logged: 5:00:14 PM

Error fetching alexa_media data: Error communicating with API: 'NoneType' object is not subscriptable

AND 

Logger: homeassistant.core
Source: core.py:2650
[...]

Issue #2892, opened 2025-04-10
Rivian Device causing Exception on Discovery


jrgutier:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

Rivian Alexa Device is causing the device API query to fail. 

It is confirmed to work once I delete the device. The Rivian device used to work without erroring, but something changed in the API response that is causing it to parse a NoneType.

**To Reproduce**

1. Activate Alexa Integration in Rivian
2. Restart Alexa Media Player and get no media

**Expected behavior**

1. The parser should be able to handle Rivian device.
2. The parser should handle bad data and not fail to load any entities because it encountered data it didn't expect.

**Screenshots**

 

**System details**

- Home Assistant version: 2025.4.1
- alexa_media version (from `const.py` or HA startup log): 5.6.0
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log): 1.29.5
- Is Amazon [2SV via Authenticator App](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (not 2SV OTP via SMS) enabled? YES
- Amazon Domain: amazon.com

**Debug Logs (alexa_media & alexapy)**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

```
2025-04-09 20:40:38.256 DEBUG (MainThread) [custom_components.alexa_media] Alexa entities have been loaded. Prepared for discovery.
2025-04-09 20:40:38.256 ERROR (MainThread) [custom_components.alexa_media] Error fetching alexa_media data: Error communicating with API: 'NoneType' object is not subscriptable
2025-04-09 20:40:38.256 DEBUG (MainThread) [custom_components.alexa_media] Finished fetching alexa_media data in 0.558 seconds (success: False)
```

**Additional context**

keenanj123:
My AMP was doing the same thing no devices and I was getting connection lost from home assistant in general not sure it it was related.  Once I removed the Rivian I got my devices back.  Thanks for finding this hope they get it resolved!

keenanj123:
This seems to have fixed the issue with my Rivian!

keenanj123:
[...]

PR #2894, opened 2025-04-12
fix: handle parsing alexa entities when network_details is none

natekspencer:


natekspencer:
I believe this should resolve the following as they contain the same error I was getting:
https://github.com/alandtse/alexa_media_player/issues/2893
https://github.com/alandtse/alexa_media_player/issues/2892

custom_components/alexa_media/alexa_entity.py
```diff
@@ -263,7 +263,9 @@ def parse_alexa_entities(network_details: Optional[dict[str, Any]]) -> AlexaEnti
     air_quality_sensors = []
     contact_sensors = []
     switches = []
-    location_details = network_details["locationDetails"]["locationDetails"]
+    location_details = (
+        network_details["locationDetails"]["locationDetails"] if network_details else {}
```
alandtse: Might be best to also do a double get of locationDetails with a default of {} in case the nested dict is malformed.

natekspencer: Done

alandtse (approved):

PR #2909, opened 2025-05-07
fix: Ensure Alexa-Compatible MP3 Output Using FFmpeg Flag

meirlo:
According to the [Alexa SSML documentation](https://developer.amazon.com/en-US/docs/alexa/custom-skills/speech-synthesis-markup-language-ssml-reference.html#h3_converting_mp3), the `-write_xing 0` flag must be used when generating MP3 files to ensure compatibility with Alexa.

This change adds the necessary FFmpeg flag during MP3 creation to produce files that meet Alexa's playback requirements.


alandtse:
Thanks!

PR #2939, opened 2025-07-05
Fix: Add workarounds for Alexa Guard API paywall issue

sharkyy:
Alexa Guard API calls are reportedly failing due to a new paywall. This change implements workarounds in the alexa_media_player custom component:

- Wraps Alexa Guard API calls (set_guard_state, static_set_guard_state) in `alarm_control_panel.py` with try-except blocks to catch errors.
- Logs warnings when Guard API calls fail, informing the user about potential subscription issues.
- Implements a graceful fallback for the Guard entity state. If Guard data cannot be fetched via the coordinator, or if an unknown state is received, the entity will report as 'unavailable'.
- This aims to prevent Home Assistant crashes and ensure media player entities remain stable even if Guard functionality is impaired.

### DIFF
--- a/custom_components/alexa_media/alarm_control_panel.py
+++ b/custom_components/alexa_media/alarm_control_panel.py
@@ -12,6 +12,7 @@ import logging
 from typing import List, Optional
 
 from alexapy import hide_email, hide_serial
+from alexapy.errors import AlexapyLoginError, AlexapyPyotpError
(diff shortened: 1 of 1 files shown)


sharkyy:
Not tested good enough, will come back after further tests

danielbrunt57:
Amazon's `/api/phoenix` is now fully deprecated.  This prevents AMP from getting any information about ancillary devices in the Alexa network, including Alexa Guard.

PR #2989, opened 2025-08-07
fix: replace deprecated MediaPlayerState.STANDBY

figorr:
The deprecated enum member MediaPlayerState.STANDBY was used from alexa_media. It will be removed in HA Core 2026.8.0. Use MediaPlayerState.OFF or MediaPlayerState.IDLE

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated media player behavior to show as "Off" instead of "Standby" when not active or playing.

### DIFF
--- a/custom_components/alexa_media/media_player.py
+++ b/custom_components/alexa_media/media_player.py
@@ -1021,7 +1021,7 @@ class AlexaClient(MediaPlayerDevice, AlexaMedia):
             return MediaPlayerState.PAUSED
         if self._media_player_state == "IDLE":
             return MediaPlayerState.IDLE
-        return MediaPlayerState.STANDBY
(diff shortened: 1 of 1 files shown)


juspitt:
I wouldn't use .OFF use .IDLE instead it could cause problems

If you use OFF
- HA will think the device is powered down.
- Some automations or UI elements won’t let you play media without “turning it on” first.
- Voice commands through the integration might not be sent until you “wake” the device.

If you use IDLE
- HA knows the device is on, just not doing anything.
- You can immediately start playback, send TTS, etc.

figorr:
Thank you @juspitt.

File updated with .IDLE.

You are right ... it looks like voice commands were broken by using .OFF

PR #3021, opened 2025-09-26
fix: Model ID for Echo Dot (Gen4)

alams154:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the displayed model name for an Echo Dot (Gen4) device that was previously labeled as “Echo Dot Clock (Gen4)”.
  * Ensures the correct product name appears in the UI, reducing confusion between the standard and clock variants.
  * No behavioral or functional changes; this is a label update only.


alams154:
Gateway Timeout (504) for Coverage test, not sure how to re-run

PR #3127, opened 2025-11-14
fix: Reduce unnecessary API calls

nao-pon:
Reduce unnecessary API calls by avoiding calls to `AlexaAPI.get_network_details` in cases where you don't need devices connected to your Echo device.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network information retrieval by requiring extended entity discovery to be enabled during updates.


nao-pon:
@alandtse So quickly! Thanks! :+1:

PR #3140, opened 2025-11-17
revert: "fix: reduce unnecessary API calls (#3127)"

nao-pon:
This reverts commit 031e5ea7bf1f87d284769e274a65ff5a8b811811.

The impact was greater than I had initially anticipated, so I am reverting.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal logic for controlling network data retrieval to read configuration directly from data storage.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -433,13 +433,12 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         ].values()
         auth_info = hass.data[DATA_ALEXAMEDIA]["accounts"][email].get("auth_info")
         new_devices = hass.data[DATA_ALEXAMEDIA]["accounts"][email]["new_devices"]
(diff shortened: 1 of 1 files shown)


alandtse:
Please be very sure of what you're doing. We should not be committing and then reverting on a regular basis. If code is not tested or ready, leave the PR in a draft state please.

nao-pon:
Yes, I'm sorry. There was an issue where API access errors occurred and entities were not displayed, so I made the change to avoid unnecessary API calls as much as possible. However, since I don't have a compatible device, I was unable to verify whether sensors such as temperature were malfunctioning.

Also, the issues I was initially concerned about were resolved by changes to #3138 and alexapy, so I have decided to revert them.

I will be more careful and cautious when making corrections in the future.

danielbrunt57:
@naopon I'm closing this PR as it is now somewhat outdated since newer PR's also address reverting [fix: reduce unnecessary API calls(#3127)](https://github.com/alandtse/alexa_media_player/pull/3127)

PR #3143, opened 2025-11-19
fix: error communicating with api notifications

danielbrunt57:
This requires `alexapy` [MR 425!](https://gitlab.com/keatontaylor/alexapy/-/merge_requests/425) to fix the error but also makes `process_notifications()` handling more robust with additional logging


nao-pon:
This PR branch does not originate from the current alandtse:dev branch. I recommend rebasing it onto the current alandtse:dev branch.

PR #3150, opened 2025-11-20
fix: Make `process_notifications()` handling more robust while adding additional debug logging

danielbrunt57:
style: The PR makes `process_notifications()` more informative with additional debug logging.

fix: This code checks if `raw_notifications` is 'truthy' i.e. Does the variable contain meaningful data to process?
Previously the logic only checked if raw_notifications is 'None' and would allow an empty response which it would then try and process.

The source of the current error being observed though is in:
`alexapy` [MR 425!](https://gitlab.com/keatontaylor/alexapy/-/merge_requests/425)'
It's summary is:
Modify 'async def get_notifications()' to prevent "Error communicating with API: 'notifications'" when the returned JSON from /api/notifications doesn't always contain a "notifications" key (for errors, rate limits, login issues, etc.) and indexing it with ["notifications"] then blows up resulting in the _catch_all_exceptions decorator turning it into the general error string being seen.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate notification timestamp handling; attaches computed notification date/time for alarms.
  * Normalizes "MusicAlarm" to "Alarm", skips notifications missing device identifier, improves alarm dismissal detection and fires dismissal events.
  * Clearer logging when no notifications are retrieved and detailed per-account/per-device update counts.

* **Refactor**
  * Notifications now stored in a structured per-account/per-device/per-type/per-id layout for more consistent updates and reporting.


danielbrunt57:
Superseded by PR #3176

PR #3160, opened 2025-11-23
fix: notifications API None handling

briankeefe:
## Description
Fixes a TypeError that occurs when Amazon's notifications API endpoint returns a 400 error, causing `raw_notifications` to be `None`.

## Problem
When the `/api/notifications` endpoint returns a 400 Bad Request, the integration crashes during the debug logging statement at line 786 when attempting to call `len(raw_notifications)` on a `None` value.

**Error:**
```
Error fetching alexa_media data: Error communicating with API: object of type 'NoneType' has no len()
```

## Solution
Added a conditional check to safely handle `None` values by returning 0 when `raw_notifications` is `None`:
```python
len(raw_notifications) if raw_notifications else 0
```

## Impact
- Fixes integration setup failures when Amazon's notifications API is unavailable or returns errors
- Allows devices to be discovered and entities created even when notifications endpoint fails
- No functional changes to actual notification handling logic

## Testing
- Tested with Home Assistant where notifications API was returning 400 errors
- Integration successfully loaded and created all device entities after fix
- No impact on normal operation when notifications API works correctly

## Related
This is a defensive fix that allows the integration to gracefully handle API errors rather than crashing during debug logging.

## Bug Fixes
* Improved error handling in notification processing to prevent potential runtime failures.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -783,7 +783,7 @@ async def setup_alexa(hass, config_entry, login_obj: AlexaLogin):
         _LOGGER.debug(
             "%s: Updated %s notifications for %s devices at %s",
             hide_email(email),
-            len(raw_notifications),
+            len(raw_notifications) if raw_notifications else 0,
             len(notifications),
             dt.as_local(
                 hass.data[DATA_ALEXAMEDIA]["accounts"][email]["notifications"][


SteveCInVA:
Can confirm this resolves the issue with Alexa devices reporting as not visible.

12oclocker:
As of 11-25-2025 I can also confirm this resolves the issue with Alexa devices missing. This pull should be applied right away to restore functionality.

ftclray:
works, simple fix.. thanks to all!

P3TACCO:
This fix resolved the Alexa Devices and the notify.alexa service missing, thank you.

I'm still learning and finding the right line to modify was a bit of a challenge given the instructions above.

For some reason, when searching in the "custom_components/alexa_media/__init__.py" file, the "process_notifications" function wasn't being found.  (fat fingered the search maybe - might have had a space in there?)

Once I found the function about 2/3rds the way down, I was a bit confused with where to insert to fix.  I saw several references to "raw_notifications". 
Finally realized it's a cut/paste replacement.

I replaced "len(raw_notifications)"
with
"len(raw_notifications) if raw_notifications else 0"

Which once found and completed seemed logical and straightforward, just a bit confusing to locate.

I'm going to attach the modified file to save any other newbs the hassle.  :-)

[__init__.py](https://github.com/user-attachments/files/23697569/__init__.py)

PR #3190, opened 2025-12-01
fix: add smart_switch type to alexa_entity

ricfio:
## Summary by CodeRabbit

* **New Features**
  * Added support for smart switches, expanding the range of compatible devices that can be controlled through the integration.


alandtse:
Does this actually change anything?

ricfio:
> Does this actually change anything?

Yes, it changes the type definition.
I added smart_switch because the function returns a smart_switch field, but the AlexaEntities TypedDict did not declare it.
Without this key, type checkers and IDEs report the returned value as inconsistent with the declared type.
This update makes the return type accurate and prevents static-analysis warnings.

alandtse:
@coderabbitai please confirm the changed code actually should result in changes to application behavior.

Issue #3177, opened 2025-11-27
Translation Error: The intl string context variable "otp_secret" was not provided to the string "{otp_secret} inválida. {message}"


danielbrunt57:
> I just received the update to version 5.8.0.
> I updated, and a new error appeared.
> Has anyone else received this error, or is it exclusive to the Brazilian Portuguese language?
> 
> <img width="1179" height="232" alt="Image" src="https://github.com/user-attachments/assets/9b5c01b1-7f17-460f-8d9c-7507be24c7f4" />
>  

 _Originally posted by @wesleyRaposo in [#3137](https://github.com/alandtse/alexa_media_player/issues/3137#issuecomment-3568792079)_

danielbrunt57:
@wesleyRaposo,

There are two things at play here.
1. You're entering an invalid Authenticator App key thus triggering the `2fa_key_invalid` handling.
2. There's a bug in the `2fa_key_invlalid` setup in `config_flow.py` where `{otp_secret}` is not being passed to the routine.

Thanks for revealing the bug (which is why I created a new issue)!

Replace your `pt-BR.json` in the `translations` folder with this revised file: [pt-BR.json](https://github.com/user-attachments/files/23808567/pt-BR.json)
which will change the messages to be more informative.

Also replace `config_flow.py` with this revised version: [config_flow.py](https://github.com/user-attachments/files/23808495/config_flow.py)
which addresses the error that's occurring.

Restart HA, reload the browser window to refresh cached files and retry configuring the integration with the same values you were giving it before.  The `2fa_key_invalid` window should now open properly

wesleyRaposo:
@danielbrunt57 

Shortly before its publication, I had already discovered the problem.
In this case, there are two files for the Brazilian Portuguese language:

`pt_BR` and
`pt-BR`.

Since the file for Portuguese from Portugal is named `pt_PT`, I was led to adjust only the `pt_BR` file.

By replicating what I did to the `pt_BR` file to the `pt-BR` file, the error stopped occurring.

If you want, use the file I sent you. I believe the translation is satisfactory.

<img width="347" height="940" alt="Image" src="https://github.com/user-attachments/assets/bcffcb56-5bfb-47bf-81a4-1fcff06c645c" />

wesleyRaposo:
It would be interesting to see which file is redundant and delete it.

danielbrunt57:
> It would be interesting to see which file is redundant and delete it.

[...]

Issue #3169, opened 2025-11-24
Validation of translation placeholders


blue-bean:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

New error message with v. 5.8.0

**To Reproduce**

1. Install AMP 5.8.0
2. See error message

**Expected behavior**

No error message

**System details**

- Home Assistant version: 2025.11.3
- alexa_media version (from `const.py` or HA startup log): 5.8.0
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log):
- Is Amazon [2SV via Authenticator App](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (not 2SV OTP via SMS) enabled?  (y/n): y
- Amazon Domain: amazon.de

**Debug Logs (alexa_media & alexapy)**
-

**Additional context**

Error message:
Logger: homeassistant.helpers.translation
Quelle: helpers/translation.py:285
Erstmals aufgetreten: 05:52:37 (1 Vorkommnis)
Zuletzt protokolliert: 05:52:37

Validation of translation placeholders for localized (de) string component.alexa_media.config.error.2fa_key_invalid failed: (set() != {'otp_secret', 'message'})

danielbrunt57:
This will be solved automatically once **_Lokalise_** kicks in and translates the new `strings.json` file which will result in a new AMP release.  For now as a temporary workaround:

1. In `config/custom_components/alexa_media/`, copy `strings.json` to `translations/de.json` - This will give you English descriptions.
2. Edit `de.json` to match the changes in `strings.json` translating any new keys to your language.

Then restart HA. Shift-Reload your browser page to refresh the cache.

The above applies to any other language translation.

PR #3191, opened 2025-12-01
fix: remove 2fa_key_invalid placeholder

ricfio:
- FIX [#3185](https://github.com/alandtse/alexa_media_player/issues/3185)
- FIX [#3177](https://github.com/alandtse/alexa_media_player/issues/3177)
- FIX [#3169](https://github.com/alandtse/alexa_media_player/issues/3169)
- FIX italian translation for 2fa_key_invalid error

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated two-factor authentication error messages across all supported languages.

### DIFF
--- a/custom_components/alexa_media/strings.json
+++ b/custom_components/alexa_media/strings.json
@@ -12,7 +12,7 @@
       "invalid_auth": "Login was not successful. Please double-check your email, password, and Authenticator key.",
       "oauth_error": "Could not complete OAuth login. Please try again.",
       "invalid_url": "URL is invalid: {message}",
-      "2fa_key_invalid": "{otp_secret} is invalid.\n{message}",
+      "2fa_key_invalid": "{otp_secret} is invalid",
       "unable_to_connect_hass_url": "Unable to connect to Home Assistant Local URL. Please check the URL under Settings > System > Network > Home Assistant URL > Local network",
       "unknown_error": "Unknown error: {message}"
     },


alandtse:
Thanks. Only strings.json needs to be modified. Others are autogenerated on build so all changes will be wiped. Please keep only strings.json.

ricfio:
> Thanks. Only strings.json needs to be modified. Others are autogenerated on build so all changes will be wiped. Please keep only strings.json.

Ah ok, sorry, I didn’t know that… now I’ve changed only the placeholder in strings.json.

alandtse:
@danielbrunt57 I believe your change removed the message. Please confirm it's not actually used and approve.

danielbrunt57:
> @danielbrunt57 I believe your change removed the message. Please confirm it's not actually used and approve.

I approve.

I've confirmed that the current `config_flow.py` doesn't pass the "message" placeholder thus creating an error since the current `strings.json` expects it.  I don't see any other PR's from myself or others that reinstates a "message" placeholder so this PR is required.

I'm working up a different scheme which checks the key the user input and will pass different messages depending on what they input (TOTP code, incorrectly formatted 2SV key, etc.) so they are better informed of the mistake they made.  However, to permit translation of different messages, I'll need to add additional translatable `2fa_key_invalid` keys with different messages embedded (rather than placeholders) each will use just the "otp_secret" they input.

alandtse:
I thought we use to have a message, are you saying it's totally gone now? @danielbrunt57 If it makes sense to drop the message that's fine, but I don't want to lose info if it was inadvertent.

danielbrunt57:
> I thought we use to have a message, are you saying it's totally gone now? @danielbrunt57 If it makes sense to drop the message that's fine, but I don't want to lose info if it was inadvertent.

Nope.  In the past there were no placeholders at all.  This is v5.7.12 `strings.json`: `"2fa_key_invalid": "Invalid Authenticator App key for Amazon 2SV"`.

I attempted to improve what the user saw for an error message but with the myriad of conflicting PR's, my `strings.json` PR that added {`otp_secret}` and `{message}` got messed up.  This PR corrects my error and I will regroup and retry as I now have a better approach in mind which will use multiple different "2fa_key_invalid" strings (i.e. "2fa_key_invalid_1", "2fa_key_invalid_2", etc., or something like that) and depending on whether they entered a 6-digit OTP code or an improper key: too short/too long/fails to generate OTP code, etc. (if I can think of any other scenario) will present an informative error message in their language.  The message will not be passed but by using different strings, the text can be different and will be translated with the goal being to make the error messages more intelligent, informative & easy to read which hopefully will result in fewer support issues!

ysfchmsdn:
I still cannot complete the login while adding the *Alexa Media Player* integration in Home Assistant.  
During the configuration flow, the integration keeps rejecting my app-based OTP.  
The dialog shows the message `<OTP_CODE> is invalid` (see screenshot).

I can reproduce this every single time.
image

danielbrunt57:
> I still cannot complete the login while adding the _Alexa Media Player_ integration in Home Assistant. During the configuration flow, the integration keeps rejecting my app-based OTP. The dialog shows the message `<OTP_CODE> is invalid` (see screenshot).
> 
> I can reproduce this every single time. image

Read the [installation & configuration wiki](https://github.com/alandtse/alexa_media_player/wiki/Configuration#2sv-authenticator-app-key---required)

PR #3240, opened 2025-12-29
chore: do AI suggested changes #3239

SSSM0602:
- Added truthiness check for binary_entities (helpers.py)
- Added period to docstring for consistency (helpers.py)
- Fix to type checking in safe_get (helpers.py)
- Added defensive truthiness for light_entities (light.py)
- Updated timer type normalization logic (media_player.py)
- Updated dependency version constraints (pyproject.toml, manifest.json)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented errors during extended device discovery when certain device types are absent.

* **Improvements**
  * Stricter type handling for defaults and more robust media timer/value processing to reduce edge-case failures.

* **Dependencies**
  * Tightened dictor version constraint to ensure compatible releases.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5793,4 +5793,4 @@ ifaddr = ">=0.1.7"
 [metadata]
 lock-version = "2.1"
 python-versions = ">=3.13,<4.0"
-content-hash = "6bbea25af061a89c40014b78c5cb174cad26162232a2d816c1b468564e109264"
+content-hash = "7a438b877a183816224eeef729c9471465bc924a4c97582ca798d078b60020e4"


alandtse:
@SSSM0602 thanks. Please generate an appropriate poetry lock given the change to pyproject.

alandtse:
@nao-pon @SSSM0602 there's a test failure. Is it a bad test or did the suggested edits break backwards compatibility.

danielbrunt57:
@nao-pon, ChatGPT provided me this:

> Your `safe_get` is almost certainly doing a **strict type match** (e.g. `type(val) is type(default))`, so when the mocked value is `True` and the default is `0` (an `int`), it rejects `True` (because `type(True) is bool`, not `int`) and falls back to the default `0`.
> 
> That’s why you see:
> - expected: `True`
> - got: `0` → so the function returned the default
> 
> #### Fix
> Use `isinstance()` except when the expected type is `bool`, because `isinstance(1, bool)` is `False` (good) but `isinstance(True, int)` is `True` (also good). The only trap is the other direction: `isinstance(1, int)` is `True` but you _don’t_ want `1` accepted when expecting `bool`.
> 
> So the rule is:
> - If default is `bool`: require actual `bool (type(val) is bool)`
> - Otherwise: allow subclasses via `isinstance(val, expected_type)`
> 
> Here’s the drop-in pattern:
> ```
> def _type_ok(val, default):
>     expected = type(default)
>     if expected is bool:
>         return type(val) is bool   # reject 0/1
>     return isinstance(val, expected)  # allow bool when expecting int, etc.
> ```
> Then in `safe_get`, wherever you currently do the type check, replace it with `_type_ok(...)`.
> 
> #### Example patch (minimal)
> If you currently have something like:
> ```
> if type(value) is type(default):
>     return value
> return default
> ```
> Change to:
> ```
> expected = type(default)
> if expected is bool:
>     return value if type(value) is bool else default
> return value if isinstance(value, expected) else default
> ```
> #### Why this makes your tests behave
> - default 0 (int), value True (bool):
>   - expected is int → isinstance(True, int) ✅ → returns True
> - default False (bool), value 1 (int):
>   - expected is bool → type(1) is bool ❌ → returns default False

nao-pon:
Thanks! all, I fixed safe_get().

nao-pon:
@SSSM0602 Could you please merge the dev branch and regenerate `poetry.lock`?

SSSM0602:
I just regenerated the `poetry.lock` file. 
Just out of curiosity, is it worth regenerating if all that changes is the content hash at the end of the file? I'm not entirely familiar with poetry but to my understanding it handles all the project dependencies.  

This was my first open source PR contribution so I appreciate all the comments and info :)

alandtse (approved):

Issue #3206, opened 2025-12-05
Timeout error occurred accessing AlexaAPI: An exception of type CancelledError occurred. Arguments: ()


Dzoker2:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

Getting this error in the log:

This error originated from a custom integration.

Logger: alexapy.helpers
Source: custom_components/alexa_media/__init__.py:869
integration: Alexa Media Player (documentation, issues)
First occurred: 8:30:59 AM (1 occurrence)
Last logged: 8:30:59 AM

alexaapi.get_customer_history_records((<alexapy.alexalogin.AlexaLogin object at 0x7f1efc196cf0>,), {'max_record_size': 10}): Timeout error occurred accessing AlexaAPI: An exception of type CancelledError occurred. Arguments: ()

**To Reproduce**

1. Go to Logs
2. See error

**Expected behavior**

Not to see the error

**System details**

- Home Assistant version: 2025.12
- alexa_media version (from `const.py` or HA startup log): 5.7.5
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log): 5.8.1
- Is Amazon [2SV via Authenticator App](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (not 2SV OTP via SMS) enabled?  (y/n): Yes
- Amazon Domain: amazon.com

**Debug Logs (alexa_media & alexapy)**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).
Don't have anything else

**Additional context**

blue-bean:
@danielbrunt57
 PR https://github.com/alandtse/alexa_media_player/pull/3241 didn't fix this error.
With v. 5.11.0, 5.12.0 and also 5.12.1 the error message still occurs in the log.
In 5.10.0 there is no such error message.

danielbrunt57:
It was actually [PR #3313](https://github.com/alandtse/alexa_media_player/pull/3313) that should have fixed this.

danielbrunt57:
@blue-bean Could you please provide a debug log trace up to and including the instance of the error, thanks.

blue-bean:
Here the log

[...]

Issue #3235, opened 2025-12-25
[Bug]: Error while trying to call a routine through an automation


rodrigogbs:
### Describe the bug

I have an automation that calls a routine on my alexa device:

```
target:
  entity_id: media_player.rgbs_echo_show_8_escritorio
data:
  media:
    media_content_id: Iniciar Som Ambiente
    media_content_type: routine
    metadata: {}
action: media_player.play_media
```

It used to work for years, but now I am facing thing error on log while trying to execute it:

```
Logger: alexapy.helpers
Source: runner.py:289
First occurred: December 23, 2025 at 10:00:00 PM (78 occurrences)
Last logged: 10:37:01 AM

Resulting JSON (URL: https://alexa.amazon.com/api/behaviors/v2/automations?limit=1000), invalid type at 'None': expected dict, got list
alexaapi.get_device_preferences((<alexapy.alexalogin.AlexaLogin object at 0x7fb2640479d0>,), {}): Timeout error occurred accessing AlexaAPI: An exception of type CancelledError occurred. Arguments: ()
```

And if I go directly to the amazon routines and play it, it works perfectly.

### To Reproduce

1. Create a routine on your alexa account named "Iniciar Som Ambiente" with an action like change de volume and start playing a radio station
2. Create and automation on your HA instance with the code above and execute it

### Expected behavior

Working!

### Versions

Installation method
Home Assistant OS
Core
2025.12.4
Supervisor
2025.12.3
Operating System
16.3
Frontend
20251203.3

Alexa Media Player: Version 5.8.2

### Is Amazon 2SV via Authenticator App enabled?

Yes

### Amazon Domain

amazon.com (US)

### Debug Logs (alexa_media & alexapy)

```text

```

djdevil:
Yes also with last version no routine work but with v5.8.1 it work

rodrigogbs:
Problem persists on 5.8.3 version.

nao-pon:
The fix for this issue requires alexapy 1.29.14. AMP 5.8.3 uses alexapy 1.29.13.

Change your manifest.json to use alexapy 1.29.14, or use the unofficial edge version.

[![Open your Home Assistant instance and open a repository inside the Home Assistant Community Store.](https://my.home-assistant.io/badges/hacs_repository.svg)](https://my.home-assistant.io/redirect/hacs_repository/?owner=nao-pon&repository=alexa_media_player&category=integration)

Norsagir:
[...]

Issue #3222, opened 2025-12-09
[Bug]: Routines, via alexa media play (specific media) talk to the device from HA - no longer working


stuartrozier:
### Describe the bug

hello, your integration is brilliant - i use it to call requests from my alexa device to control ha via voice. but within the last week or so, using HA to send voice request to echo device, under "media_content_type: routine" has stopped working, 'custom' requests still work, but routines have stopped. these did work flawlessly, but ive been getting the following error recently, 

error located in Log file .... 

"Logger: alexapy.helpers
Source: runner.py:289
First occurred: 18:03:18 (2 occurrences)
Last logged: 18:26:13

Resulting JSON (URL: https://alexa.amazon.co.uk/api/behaviors/v2/automations?limit=1000), invalid type at 'None': expected dict, got list"

many thanks in advance, let me know if you need any more information as im new to reporting errors on this integration, 
stuart

### To Reproduce

action: media_player.play_media
target:
  device_id: b1caa33322207ab258b0eab1fbda10af
data:
  media:
    media_content_id: turn the T V off
    media_content_type: routine

### Expected behavior

action: media_player.play_media
target:
  device_id: b1caa33322207ab258b0eab1fbda10af
data:
  media:
    media_content_id: turn the T V off
    media_content_type: routine

### Screenshots

![Image](https://github.com/user-attachments/assets/51326fbb-8111-4b6d-82d4-6d59c0eb5dc7)

### Versions

HA ver 2025.12.2
im not what version of alexa it is, but its the latest as ui update daily

### Is Amazon 2SV via Authenticator App enabled?

Yes

### Amazon Domain

amazon.co.uk (UK)

### Debug Logs (alexa_media & alexapy)

```text

```

nao-pon:
see https://github.com/alandtse/alexa_media_player/issues/3217

Issue #3217, opened 2025-12-07
[Bug]: media_content_type: routine is not working


GareFPV:
### Describe the bug

**Script:**
```
sequence:
  - data:
      media:
        media_content_id: Alexa Büro Musik
        media_content_type: routine
        metadata: {}
    target:
      entity_id: media_player.alexa_buro
    action: media_player.play_media
mode: single
alias: Alexa Büro Musik
description: ""
```

Home Assistant Log:
`Resulting JSON (URL: https://alexa.amazon.de/api/behaviors/v2/automations?limit=1000), invalid type at 'None': expected dict, got list`

Everything else works perfectly, only the routines after the last update no longer work.

### To Reproduce

Script see above

### Expected behavior

That Alexa's routine starts

### Versions

Alexa Media Player 5.8.2

Home Assistant
Core 2025.12.1
Supervisor 2025.12.2
Operating System 16.3
Frontend 20251203.1

### Is Amazon 2SV via Authenticator App enabled?

Yes

### Amazon Domain

amazon.de (Germany)

### Debug Logs (alexa_media & alexapy)

```text

```

dahadd:
downgrade to 5.8.1 works

ahahamyan:
Had the same exact issue and had to downgrade to 5.8.1 for it to work again.

michguy2000:
Also having the same issue after updating to 5.8.2

**UPDATE**: I did the same as "ahahamyan" and downgraded to 5.8.1.   HA and Alexa are working again.

lucazade:
same here using a routine:

Registratore: alexapy.helpers
Fonte: runner.py:289
Resulting JSON (URL: https://alexa.amazon.it/api/behaviors/v2/automations?limit=1000), invalid type at 'None': expected dict, got list

nao-pon:
The root cause of this issue is alexapy. I've created a MR ([fix: for retrieving automation list from api/behaviors/v2/automations](https://gitlab.com/keatontaylor/alexapy/-/merge_requests/435)) to fix it. If you're interested in early testing, try my edge branch. You can download the edge version by adding it as a custom repository in HACS. Then restart HA.

- https://github.com/nao-pon/alexa_media_player

thexperiments:
Seems to be merged, any timeline when the dependency will be updated? Or should I also downgrade for now?

danielbrunt57:
> Seems to be merged, any timeline when the dependency will be updated? Or should I also downgrade for now?

[...]

PR #3241, opened 2025-12-30
fix: bump alexapy to 1.29.14

danielbrunt57:
#### `alexapy` version 1.29.14 should fix the following issues: 
Error while trying to call a routine through an automation #3235 
Routines, via alexa media play (specific media) talk to the device from HA - no longer working #3222
media_content_type: routine is not working #3217
Lost timer functionality #3216
Timeout error occurred accessing AlexaAPI: An exception of type CancelledError occurred. Arguments: () #3206
Timer, alarm, and reminder sensors are in 'unknown' state #3159

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -283,14 +283,14 @@ tzdata = ">=2024.1"
 
 [[package]]
 name = "alexapy"
-version = "1.29.13"
+version = "1.29.14"
 description = "Python API to control Amazon Echo Devices Programmatically."
 optional = false
 python-versions = "<4,>=3.11"
 groups = ["main"]
 files = [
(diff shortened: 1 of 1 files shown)


blue-bean:
@danielbrunt57 
This PR didn't fix the error described in https://github.com/alandtse/alexa_media_player/issues/3206
With v. 5.11.0, 5.12.0 and also 5.12.1 the error message still occurs in the log.
In 5.10.0 there is no such error message.

danielbrunt57:
> @danielbrunt57 This PR didn't fix the error described in #3206 With v. 5.11.0, 5.12.0 and also 5.12.1 the error message still occurs in the log. In 5.10.0 there is no such error message.

It was actually [PR #3313](https://github.com/alandtse/alexa_media_player/pull/3313) that should have fixed #3206.

PR #3244, opened 2025-12-30
chore!: deprecate yaml config

danielbrunt57:
BREAKING CHANGE:
YAML config of `alexa_media` is fully deprecated and will now raise an error requiring users to remove it from their configuration.

## Summary by CodeRabbit

* **Breaking Changes**
  * YAML-based configuration for the Alexa Media integration is no longer supported. If present, Home Assistant will log an error and the integration will not migrate YAML entries; remove the alexa_media entry from configuration.yaml and restart to proceed with UI setup.

* **Documentation**
  * Clarified deprecation notice and updated setup instructions to direct users to configure via the UI.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -162,7 +162,7 @@ async def async_setup(hass, config):
             "YAML configuration of Alexa Media Player is no longer supported. "
             "Please remove `alexa_media` from your configuration, "
             "restart Home Assistant and use the UI to configure it instead. "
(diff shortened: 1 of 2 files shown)


alandtse:
Given it was already deprecated, no need to mark this breaking.

alandtse (approved):

danielbrunt57:
> Given it was already deprecated, no need to mark this breaking.

One might argue that since the previous code didn't actually prevent or break YAML config but issued warnings only, this is now officially the breaking change so I chose to err on the side of caution and label this change as breaking.

PR #3285, opened 2026-01-12
fix(switch): define hue_emulated_enabled before use

jleinenbach:
The variable `hue_emulated_enabled` was used in the Smart Plug device creation logic but was never defined, causing a NameError when users had Smart Plug devices with CONF_EXTENDED_ENTITY_DISCOVERY enabled.

This bug was introduced when the Smart Switch support was added, where the code checks if a switch is a Hue v1 device that might conflict with the emulated_hue integration. Without the variable definition, users would see:

  NameError: name 'hue_emulated_enabled' is not defined

The fix adds the missing variable definition that checks if 'emulated_hue' is present in the Home Assistant components configuration.

Also adds comprehensive tests for the Smart Switch creation logic covering:
- Verification that no NameError occurs with smart switches
- Non-Hue-v1 switches created regardless of emulated_hue setting
- Hue-v1 switches skipped when emulated_hue is enabled
- Hue-v1 switches created when emulated_hue is not enabled
- No switches created when extended discovery is disabled
- emulated_hue detection logic

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where certain smart switches were incorrectly created when Hue v1 emulation was enabled, preventing potential device control conflicts.

* **Tests**
  * Added comprehensive test coverage for smart switch creation logic with various Hue emulation configurations.

### DIFF
--- a/custom_components/alexa_media/switch.py
+++ b/custom_components/alexa_media/switch.py
@@ -67,11 +67,8 @@ async def async_setup_platform(hass, config, add_devices_callback, discovery_inf
                 hide_serial(key),
             )
             raise ConfigEntryNotReady
-        if (
-            key
-            not in (
-                hass.data[DATA_ALEXAMEDIA]["accounts"][account]["entities"]["switch"]
-            )
+        if key not in (
+            hass.data[DATA_ALEXAMEDIA]["accounts"][account]["entities"]["switch"]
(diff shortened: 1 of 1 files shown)


jleinenbach:
@coderabbitai help

PR #3291, opened 2026-01-12
fix: rewrite tests with proper runtime imports to fix CI test failure

jleinenbach:
## Summary

Rewrites `test_sensor.py` and `test_helpers.py` with proper runtime imports to fix CI test failures.

## Problem Analysis

The original tests from PRs #3282 and #3288 had two issues:

1. **Incorrect test assertions**: `TestAddDevices.test_add_devices_condition_error_*` tests expected `False` but `add_devices()` returns `True` after catching exceptions (the function logs errors but still returns `True` for handled exceptions)

2. **Missing test coverage**: The original sensor tests attempted to import `AlexaMediaNotificationSensor` but didn't properly mock the dependencies, causing import chain failures in certain test execution orders

## Solution

Both test files now use proper runtime imports with `pytest-homeassistant-custom-component`:

### test_sensor.py (6 tests)
- Imports `AlexaMediaNotificationSensor` directly
- Tests `_update_recurring_alarm` method with proper mocking
- Verifies the `isoweekday()` bugfix (parentheses required)
- Tests alarm advancement logic, status handling (ON/OFF), and reminder types

### test_helpers.py (32 tests)
- Imports helper functions directly from `custom_components.alexa_media.helpers`
- Tests `add_devices` filter defaults (regression tests for #3288 bugfix)
- Tests `_existing_serials`, `is_http2_enabled`, `safe_get` functions
- Removed incorrect `ConditionErrorMessage` assertions

## Test Results

All tests pass with the CI environment:
- Python 3.13
- `poetry run pytest`
- `pytest-homeassistant-custom-component` 0.13.236

## Test Plan

- [x] Run `poetry run pytest tests/test_sensor.py tests/test_helpers.py -v` (38 passed)
- [x] Run full test suite `poetry run pytest tests` (106 passed)
- [x] Run `ruff format` and `ruff check` (all passed)
- [x] Verify tests work with same Python/HA versions as GitHub CI

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -15,7 +15,6 @@ from custom_components.alexa_media.helpers import (
     safe_get,
 )
 
-
 # =============================================================================
 # Tests for _existing_serials function
 # =============================================================================


alandtse:
We should be importing test fixtures from pytest-homeassistant-custom-component. https://github.com/alandtse/alexa_media_player/blob/dev/pyproject.toml#L24. I'm not convinced Claude has correctly identified an issue.

jleinenbach:
> We should be importing test fixtures from pytest-homeassistant-custom-component. https://github.com/alandtse/alexa_media_player/blob/dev/pyproject.toml#L24. I'm not convinced Claude has correctly identified an issue.

Claude says the issue comes from this, but I wonder why it didn't fail earlier:
sensor.py:
`
from homeassistant.const import UnitOfTemperature, __version__ as HA_VERSION
ImportError: cannot import name '__version__'
`

__version__ from homeassistant.const is no longer available.

EDIT:
... which is not correct - Claude looked it up (HA source code) and it is still available.

Nonetheless, now it works.

alandtse (approved):

Issue #3175, opened 2025-11-25
Last Called, last called summary and last called timestamp updating inconsistently.


bravotangoau:
**IMPORTANT: Please search the issues, including closed issues, and the [FAQ](https://github.com/alandtse/alexa_media_player/wiki/FAQ) before opening a new issue. The template is mandatory; failure to use it will result in issue closure.**

**Describe the bug**

I'm using the "last called timestamp" and "last called summary" to trigger automations so I can process the commands issued to Alexa. 

From the last 2 days (at least I noticed after installing 5.8.0), my automations have been failing, I.e. not triggered, because, looking at the attributes, these 3 attributes are not always updated after commands were issued to Alexa.

**To Reproduce**

1) Simplest is to create an automation that triggers when "last called timestamp" of any of alexa media player has changed.
2) With the automation open, issue a command to the Alexa of your choice. You may or may not see the trigger being triggered.
3) If it was triggered, issue another command, or go to another Alexa and issue a command. It shouldn't take more than 2 commands to see the problem.
4) Go to Developer Tools, States and search for "media_player".
5) Check the Alexa where you issued the command. It may not have updated. 

It doesn't work (trigger) most of the time, which makes it frustrating.  Otherwise I would have just put up with it. 

I noticed that sometimes, after a long while (maybe a minute), the automation is triggered but that's way too late. 

**Expected behavior**

**Screenshots**

**System details**

- Home Assistant version: 2025.11.3
- alexa_media version (from `const.py` or HA startup log): 5.8.0
- alexapy version (from `pip show alexapy` in homeasssistant container or HA startup log):
- Is Amazon [2SV via Authenticator App](https://github.com/alandtse/alexa_media_player/wiki/Configuration#enable-two-factor-authentication-for-your-amazon-account) (not 2SV OTP via SMS) enabled?  (y/n): Y
- Amazon Domain: amazon.com

**Debug Logs (alexa_media & alexapy)**
Please provide [logs](https://github.com/alandtse/alexa_media_player/wiki/FAQ#how-do-i-enable-debug-logging-for-the-component).

**Additional context**

WeVolutionNet:
Update: [Found this](https://github.com/alandtse/alexa_media_player/issues/2929) and now I feel like reinventing the wheel ;-)


Original Answer:

**TLDR:** I found that last_called is not updated reliably anymore.
[...]

PR #3301, opened 2026-01-14
feat: Improve last_called timeliness and reliability (voice-only)

danielbrunt57:
- Uses HTTP2 push events to trigger a debounced history probe for near-real-time updates
- Keeps last_called semantics voice-only
- Filters invalid customer-history entries (punctuation/junk summaries) so fallback doesn’t poison state
- No notify/TTS/announce/drop-in effects on last_called

Fixes issue #3175 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved last-seen device updates for non-HTTP2 connections to reduce stale activity info.

* **New Features**
  * Near-real-time device activity probing to surface more responsive "last seen" information; probes are debounced/throttled to limit impact.

* **Chores**
  * Enhanced connection and setup logging for clearer diagnostics.
  * Refined internal command handling, deprecations, and probe scheduling for improved responsiveness and stability.


danielbrunt57:
#### Note: An AlexaAPI change is forthcoming which will streamline this feature further by introducing a short-window helper.

`get_customer_history_records()` defaults to 24h back and 24h forward if no start/end passed. This is fine for manual polling, but for “on push event” we want a tighter lookback like 30–60 seconds.

Adding a wrapper like:
- `get_recent_voice_activity(login, lookback_ms=60_000, max_records=10, ...)`
which calls `get_customer_history_records(start_time=end-lookback, end_time=end+small_skew)` will keep payloads small potentially reducing the risk of Amazon throttling.

danielbrunt57:
I changed this to draft as I found an issue with the code
If I execute:
```
action: notify.alexa_media_kitchen_echo_dot
data:
  message: Hello
  ```
last_called doesn't update and I think it should. I know it deviates from my NodeRED flow but I believe now that is also incorrect.

### Here's what AI had to say:

Yeah — with the current meaning of last_called in AMP, that behavior is actually expected.
#### Why notify.alexa_media_kitchen_echo_dot should NOT update last_called
AMP’s last_called is (and always has been) “the last device that heard a user voice request” (historically from PUSH_ACTIVITY, now from your near-real-time customer-history probe triggered by push events).
When you do:
```
action: notify.alexa_media_kitchen_echo_dot
data:
  message: Hello
```
that is TTS initiated by Home Assistant, not a voice interaction. It does not create a VOICE_HISTORY record, so there is nothing “true” for the last_called probe to discover. In other words:
- last_called source-of-truth = Amazon voice history / voice activity
- HA TTS ≠ voice activity
- So last_called won’t change (and arguably shouldn’t)

Your Node-RED flow “felt” like it updated because alexa-remote’s websocket activity stream is broader (it can
fire on device activity that is not strictly “voice history”). That makes it great for “last active device”, but it is not the same concept as last_called.

#### What you probably want instead
You’re describing “the last Echo that HA sent something to” (or “last active device”), which is a different semantic.

Best options:

#### Option A (recommended): Add last_notified
Keep last_called pure (voice-driven), and add a separate tracking path:
- per-entity attribute: last_notified: true/false
- timestamp attribute: last_notified_timestamp
- account-level last_notified mapping for routing / sensors
- optional event: alexa_media_last_notified_event
That gives you what you want without corrupting the meaning of last_called.
#### Option B (configurable): Optional “treat notify as last_called”
If you really want notify.* to advance last_called, it should be opt-in because it will break workflows for people who expect “last voice device”.

This would look like an option:
- last_called_updates_on_notify: false (default)
- if true, after a successful notify send to exactly one device, set that device as “last_called”
But note: this will diverge from Amazon’s own history, and may cause confusing “why is last_called changing when nobody spoke?” reports.

### I'm looking for feedback on how to proceed...

**Comment by danielbrunt57** [comment]
@jleinenbach , @nao-pon , @alandtse ,
Your thoughts?

**Comment by jleinenbach** [comment]
@danielbrunt57 
I don't have much time right now, so asked Claude for you (copy/paste), I hope this helps:

## Feedback on the `last_called` Semantics Question

Thanks for the ping, @danielbrunt57! I've reviewed the PR and the conceptual question you raised. Here's my perspective:

### TL;DR
**`last_called` should remain strictly "device that heard a user voice command"** — TTS notifications from HA should NOT update it. A separate `last_notified` attribute would be a clean addition if needed.

---

### Reasoning

#### 1. The Primary Use Case is Voice-Triggered Room Context

The core value of `last_called` is enabling **room-aware automations** based on where the user *spoke*. The canonical example from the [community discussion](https://community.home-assistant.io/t/using-alexa-media-player-last-called-service/332546):

> "Alexa, turn off TV" → HA knows *which* TV because it knows *which Echo heard the command*

This pattern only works if `last_called` reflects **user intent through voice input**, not system-initiated output.

#### 2. TTS Updates Would Break Existing Automations

Consider this scenario:

User workflow:

1. User says "Alexa, what's the weather?" in the KITCHEN
2. last_called = kitchen_echo ✓
3. HA automation sends TTS doorbell notification to LIVING_ROOM echo
4. If TTS updated last_called → last_called = living_room_echo ✗
5. User says "Alexa, turn on the lights" (expecting kitchen)
6. Wrong room gets triggered!


The fundamental issue: **TTS is an outbound action initiated by HA**, not an inbound signal from the user. Mixing these two concepts in one attribute conflates "where did the user speak?" with "where did HA last send output?" — two semantically different questions.

#### 3. Separation of Concerns

From a software design perspective:

| Attribute | Semantic Meaning | Trigger |
|-----------|------------------|---------|
| `last_called` | Last device with **user voice input** | `PUSH_ACTIVITY` from Alexa |
| `last_notified` (proposed) | Last device with **HA-initiated output** | TTS/announcement sent by HA |

These are orthogonal concepts. Keeping them separate:
- Preserves backward compatibility
- Makes automations predictable
- Follows the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)

#### 4. The "Stale last_called" Problem is Real but Separate

Users have reported `last_called` becoming stale due to Amazon's API changes ([source](https://community.home-assistant.io/t/using-alexa-media-player-last-called-service/332546)). Your near-real-time probe solution (debounced HTTP2 push → voice history query) addresses this correctly by improving the *reliability* of the existing semantic, not changing what it means.

---

### Recommendation

1. **Keep `last_called` as voice-command-only** (current behavior)
2. **Optionally add `last_notified`** as a separate attribute if there's demand
3. **Document the semantic clearly** in the wiki/README to prevent future confusion

The technical implementation of debounce/throttle (0.6s/2.0s) looks reasonable for balancing responsiveness vs. API rate limits.

---

### Minor Technical Notes

- The `seconds` variable fix (de2f18f) looks correct
- Agree with CodeRabbit that broad exception handling could be narrowed
- Test coverage should be added before merging

Happy to discuss further!

**Comment by danielbrunt57** [comment]
Thanks!

I disagree with Claude though.
1. User says "Alexa, what's the weather?" in the KITCHEN
2. last_called = kitchen_echo ✓ (near-real-time update)
3. HA automation sends TTS doorbell notification to LIVING_ROOM echo ✓
4. last_called = living_room_echo ✓ (near-real-time update)
5. User says "Alexa, turn on the lights" (expecting kitchen)
6. last_called = kitchen_echo ✓ (near-real-time update)
7. Correct room gets triggered ✓ ← I can doublecheck this in real world

**Comment by jleinenbach** [comment]
Claude (quote):

You're right, I need to revise my reasoning.

### Where I was wrong

My scenario assumed the automation would run *before* the next voice command updates `last_called`. But that's not how it works:

1. User says "Alexa, turn on the lights" in KITCHEN
2. `PUSH_ACTIVITY` fires → `last_called` = kitchen_echo (near-real-time)
3. *Then* the automation executes
4. Correct room ✓

The voice command updates `last_called` **before** the triggered automation runs, so the state is always correct for voice-initiated flows. My concern was based on a flawed mental model of the event sequence.

### Where a (minor) edge case remains

The only scenario where TTS updating `last_called` *could* cause unexpected behavior is with **non-voice-triggered automations** that use `last_called`:

```yaml
# Example: Timer-based morning greeting
trigger:
  - platform: time
    at: "08:00:00"
action:
  - service: notify.alexa_media_last_called
    data:
      message: "Good morning!"
```

If a TTS went to the living room at 3 AM, the morning greeting would go there instead of where the user last spoke.
But realistically: Anyone using last_called in a timer automation probably wants "most recently active device" anyway, not strictly "where I last spoke." So even this edge case may not be a problem in practice.
Updated position
I concede the point. For the vast majority of use cases, the near-real-time update makes my concerns moot. Whether TTS updates last_called or not is more of a semantic/philosophical choice than a breaking behavioral change.
If you decide TTS should update last_called, the impact on existing automations would be minimal. A separate last_notified attribute would still be cleaner from a purist perspective, but I no longer see it as necessary.
Thanks for pushing back — always good to have assumptions challenged.

danielbrunt57:
I modified my old script `lights`:
```
sequence:
  - if: []
    then:
      - wait_for_trigger:
          - trigger: state
            entity_id:
              - sensor.alexa_media_last_called
        timeout:
          hours: 0
          minutes: 0
          seconds: 5
          milliseconds: 0
        continue_on_timeout: false
      - data: {}
        target:
          area_id: "{{ area_id(states('sensor.last_alexa')) }}"
        action: light.toggle
alias: Lights
mode: single
icon: hass:lightbulb-group
description: ""
```
and tested:
3. HA automation sent TTS "Home assistant has started" notification to LIVING_ROOM echo ✓
4. last_called = media_player.echo ✓ via `if not http2_enabled: await update_last_called(login_obj)` in `async_update_data()`
5. User says "Alexa, lights" (expecting bathroom)
6. last_called = media_player.bathroom_echo_dot ✓ (near-real-time update)
7. Correct room was triggered (bathroom) ✓

Takeaway:  `update_last_called()` _appears to_ update last_called when TTS notifications are sent but the new near-real-time update does not.
Confirmation: I sent another TTS notification to kitchen_echo_dot and the new near-real-time update _did not_ update. I then manually invoked `alexa_media.update_last_called` and `last_called` changed to `kitchen_echo_dot`.

danielbrunt57:
> Updated position
> I concede the point. For the vast majority of use cases, the near-real-time update makes my concerns moot. Whether TTS updates last_called or not is more of a semantic/philosophical choice than a breaking behavioral change.
> If you decide TTS should update last_called, the impact on existing automations would be minimal. A separate last_notified attribute would still be cleaner from a purist perspective, but I no longer see it as necessary.

I confirmed that my Node-RED published device activity for TTS but not for announce.  It also published the _last device_ from voice history when multiple entities are targeted and I'm not sure that's what we want/need.
- Node-RED updates because Amazon’s “customer history” ends up attributing the TTS to one device (often the one that effectively “finishes last” or the last device in the request list, depending on how the behavior call lands).
- My new HA synth-hook (notify.py) currently only fires for single-device sends to avoid ambiguity — so multi-target TTS won’t touch `last_called`.
If we want AMP to behave closer to Node-RED for multi-target sends, we need to pick a deterministic rule.
#### AI Recommended rule (matches my example)
When multiple targets are specified, set last_called to the last target in the user’s target list (i.e., the last device the service call requested). In your YAML, that would be media_player.echo_show_8, which matches what Node-RED reported.

#### Question: Do we want to mimic Node-Red behavior or do we set our own standard and **only** update `last_called` when a single device is targeted?

alandtse:
So I realize that the summary was an AI draft, but last called has always been intended to be based on a http push event to identify the last alexa that a user interacted with by voice. This means by design it is a single entity although it could be multiple if there are multiple accounts since more than one Alexa can exist in a room under multiple accounts.

danielbrunt57:
Thanks @alandtse .  So, thinking back to the good ole days when Amazon pushed activity, that would have been true, and I can modify it to adhere to that design intent but, what about update_last_cslked?  It currently retrieves the last voice record which is ",".  Should we fix that to adhere to the original intent "voice only"?

I'll also review to ensure it handles multiple accounts correctly, and revise the PR description to reflect the final behaviour of this fix.

alandtse:
The last voice record was always a stop gap in case the http websocket failed. If it's not good data, then it shouldn't be used anymore.

danielbrunt57:
Thanks. I will proceed then to keep last_called as voice only. 

If I add last_notified as new feature, it will need to change more files than this one does. Should I make it a new separate PR?

danielbrunt57:
@alandtse,
Re: semantics — alexapy documents `get_last_device_serial()` as “find the latest entry where Echo successfully responded,” which is broader than strictly “voice-only.”

For this PR, AMP keeps `last_called` voice-focused by ignoring customer-history entries whose `summary` contains no alphanumeric characters, preventing artifacts like “,” from overwriting state.

Separate but related: I traced the "," summary back to alexapy’s transcript concatenation logic — Amazon’s record can be empty, and alexapy’s string-building can leave a dangling comma when joining empty parts. I’m happy to open a separate alexapy PR to switch that logic to building a list of non-empty transcript fragments and ", ".join(...) them, which avoids the artifact entirely.

@jleinenbach would appreciate your thoughts on whether alexapy should treat empty transcript fragments as “no summary” rather than emitting punctuation (", ").

custom_components/alexa_media/__init__.py
```diff
@@ -794,24 +793,70 @@ async def process_notifications(login_obj, raw_notifications=None) -> bool:
...
+        if not isinstance(last_called, dict) or not last_called.get("summary"):
             try:
                 async with async_timeout.timeout(10):
                     last_called = await AlexaAPI.get_last_device_serial(login_obj)
             except TypeError:
                 _LOGGER.debug(
                     "%s: Error updating last_called: %s",
                     hide_email(email),
-                    hide_serial(last_called),
+                    repr(last_called),
```
alandtse, line 846: What is this function doing?

custom_components/alexa_media/__init__.py
```diff
@@ -1145,8 +1275,8 @@ async def http2_handler(message_obj):
                 else:
                     serial = None
 
-                if command == "PUSH_ACTIVITY":
-                    #  Last_Alexa Updated
+                if command == "PUSH_ACTIVITY":  # Deprecated
```
alandtse: If the code is dead, it should just be removed.

alandtse (changes requested):

custom_components/alexa_media/__init__.py
danielbrunt57: The Python repr() function (short for representation) returns a developer-friendly, unambiguous string representation of an object, which can often be used to recreate the object's value. This is distinct from the str() function, which is designed for a human-readable output.

hide_serial() is meant to obfuscate serial-number-like strings, not arbitrary objects.

In that except TypeError: path, last_called might be:
- None
- a non-dict object
- a dict that doesn’t have the keys hide_serial() expects
- or otherwise something that causes hide_serial() (or whatever it calls internally) to raise its own exception

That’s the worst place to have logging blow up, because you’re already handling an error.

So switching to repr(last_called) makes the debug log:
- safe for any type
- non-throwing
- still useful for diagnosing what came back / what got passed in

If we know the value can contain sensitive fields and we want to keep obfuscation, a safer pattern is:
```
hide_serial(last_called) if isinstance(last_called, str) else repr(last_called)
```
…but in this specific exception handler, the whole point was “never crash while trying to log the crash,” so repr() was the simplest robust choice.

custom_components/alexa_media/__init__.py
danielbrunt57: I have removed `if command == "PUSH_ACTIVITY":  # Deprecated`

alandtse (approved):

PR #3343, opened 2026-02-07
docs: clarify intentional permissiveness of Matter bridge applianceId handling in get_device_bridge()

danielbrunt57:
This PR revises the inline code comment to make it clearer for maintainers that the previous code change was intentional. In Amazon applianceIds, `#` only appears for bridged endpoints, and HA Matter Hub uses the `AAA_SonarCloudService_…#…` prefix. The previous regex was unnecessarily strict and risked false negatives; this helper only needs to identify the bridge, not validate the full ID format.

Addresses issue #3332 

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation to clarify behavior for detecting Home Assistant Matter bridged endpoints and their associated bridges.

### DIFF
--- a/custom_components/alexa_media/alexa_entity.py
+++ b/custom_components/alexa_media/alexa_entity.py
@@ -209,8 +209,12 @@ def get_device_bridge(
     if not isinstance(appliance_id, str) or "#" not in appliance_id:
         return None
 
-    # HA Matter Hub bridged endpoints are identified by applianceId prefixes
-    # of the form AAA_SonarCloudService_<bridgeId>#<childId>.
+    # This change intentionally avoids over-validating the applianceId. In Amazon device IDs, # only appears
+    # for bridged devices, and HA Matter Hub bridged endpoints consistently use the
+    # AAA_SonarCloudService_…#… prefix. The previous regex added unnecessary constraints and risked false
+    # negatives; the current logic reflects the actual invariant.
+    # This function is not a general applianceId validator; it only answers “does this look like an HA Matter-
+    # bridged endpoint, and if so, what is its bridge?”
     bridge_id, _sep, _child = appliance_id.partition("#")
 
     if not bridge_id.startswith("AAA_SonarCloudService_"):


custom_components/alexa_media/alexa_entity.py
```diff
@@ -209,8 +209,12 @@ def get_device_bridge(
     if not isinstance(appliance_id, str) or "#" not in appliance_id:
         return None
 
-    # HA Matter Hub bridged endpoints are identified by applianceId prefixes
-    # of the form AAA_SonarCloudService_<bridgeId>#<childId>.
+    # This change intentionally avoids over-validating the applianceId. In Amazon device IDs, # only appears
+    # for bridged devices, and HA Matter Hub bridged endpoints consistently use the
+    # AAA_SonarCloudService_…#… prefix. The previous regex added unnecessary constraints and risked false
```
alandtse, line 214: There's no need to clutter with past history in comments.  Just state simply what is needed and why without comparing to code that doesn't exist anymore. If you're trying to appease the bot, that's not worth a commit.

alandtse (changes requested):
Comments need to be for people, not bots.

custom_components/alexa_media/alexa_entity.py
danielbrunt57: I was trying to appease the bot generated "Critical Issue Created:" #3332 which was triggered by your comment and assigned to me:
> @codera**itai I'm sorry, I meant to say the identified ai comments, not the issues. Can you please unassign daniel from those issues and then review comments to create new issues for critical comments on this PR? I just want to identify any blocking comments from the review vs deferable comments.

IMO, the original code comment was satisfactory.

alandtse: Then close the issue if you think the bot is wrong. No need to do a random PR if the AI is being dumb.

alandtse:
Not needed per discussion.

PR #3363, opened 2026-02-14
fix: increase `get_customer_history_records()` timeout

danielbrunt57:
Some users are occasionally getting:
```
alexaapi.get_customer_history_records((<alexapy.alexalogin.AlexaLogin object at 0x7f1efc196cf0>,), {'max_record_size': 10}): Timeout error occurred accessing AlexaAPI: An exception of type CancelledError occurred. Arguments: ()
```
- Fixes issue #3206 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of last called device detection by extending the timeout duration to better accommodate slower network responses and reduce connection failures.


alandtse (approved):

alandtse:
Thanks. I wonder if we should put the timeout as an advanced setting too.

PR #3378, opened 2026-02-18
feat: Add Echo Show 11 (Gen1) to device list

alams154:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for Echo Show 11 (Gen1) device in Alexa Media component.


alandtse (approved):
Thanks!

PR #3445, opened 2026-05-07
fix: validate cookie auth before alexapy login fallback

Kyzcreig:
## Summary

When `alexa_media` is configured with a valid cookie pickle, validate those cookies against `https://alexa.amazon.com/api/bootstrap` before falling back to `alexapy.login()`.

This avoids a current failure mode where alexapy's login path can mutate/delete otherwise valid browser cookies before the real auth check completes, causing Home Assistant to drop into the Amazon Registration / reauth flow.

## Why

Observed on Home Assistant Core 2026.4.4 with alexa_media 5.15.1 + alexapy 1.29.20:

- The same `aiohttp.CookieJar` pickle returns HTTP 200 + `authentication.authenticated=true` from `/api/bootstrap` when tested directly.
- Calling `login.login(cookies=await login.load_cookie())` can still fail and trigger the credential/OAuth login fallback.
- During that fallback, the cookie pickle can be overwritten with a reduced/invalid jar and the integration goes to `setup_error`.

This patch performs a cheap non-redirecting bootstrap check first. If Amazon says the cookie session is authenticated for the configured email, it finalizes the login and skips the fragile fallback path. If the check fails, existing behavior is preserved.

## Test plan

- `python3 -m py_compile custom_components/alexa_media/__init__.py`
- Tested manually on a live HA installation: after applying the equivalent patch and restoring a valid pickle, `alexa_media` moved from `setup_error` to `loaded`; `notify.alexa_media_*` services came back.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Alexa cookie-based sign-in: now validates stored cookies against the configured account email before completing login, correctly marks successful logins, records login identifiers, and continues normal post-login checks.
  * Enhanced error handling: if cookie validation fails or errors occur, the integration reliably falls back to the standard login flow to restore access.

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -805,8 +805,13 @@ async def async_setup_entry(hass, config_entry):
                         data = loads(await response.text())
                         auth = (data or {}).get("authentication") or {}
                         customer_email = (auth.get("customerEmail") or "").lower()
-                        if auth.get("authenticated") and customer_email == email.lower():
-                            _LOGGER.debug("[BOOT] Cookie auth confirmed via /api/bootstrap")
+                        if (
+                            auth.get("authenticated")
+                            and customer_email == email.lower()
+                        ):
+                            _LOGGER.debug(
+                                "[BOOT] Cookie auth confirmed via /api/bootstrap"
+                            )
                             login.status["login_successful"] = True
                             login.customer_id = auth.get("customerId")
                             login.stats["login_timestamp"] = datetime.now()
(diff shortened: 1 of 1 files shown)


alandtse:
@danielbrunt57 please remember to hit approve if you've reviewed someone else's work and it looks good to you. You've been handling maintenance so long, if I see you've approved it, I'll likely just merge it.

danielbrunt57 (approved):

danielbrunt57:
@Kyzcreig See https://github.com/alandtse/alexa_media_player/issues/3455

PR #3457, opened 2026-05-12
fix(notify): guard targets against json.loads TypeError

### LINKED ISSUE #3453: bug: notify.py crashes on non-string targets due to unguarded json.loads() (TypeError)
## Summary

In `custom_components/alexa_media/notify.py` (lines 219–229), `json.loads(target)` is called without a type guard. Non-string target entries (e.g., dict, list, int) raise `TypeError`, which is **not** caught by the `except json.JSONDecodeError` clause. This aborts the entire notification send path for any non-string target.

The existing tests explicitly document that "non-string targets are passed through unchanged", but this contract is broken at runtime.

## Severity

**Blocker** — any non-string target in `ATTR_TARGET` will raise an unhandled `TypeError` and abort `async_send_message`.

## Proposed Fix

Add an `isinstance(target, str)` guard before calling `json.loads()` and append non-string targets directly to `processed_targets`:

```python
for target in targets:
    _LOGGER.debug("Processing: %s", target)
    if not isinstance(target, str):
        processed_targets.append(target)
        _LOGGER.debug("Processed non-string target: %s", processed_targets)
        continue
    try:
        parsed = json.loads(target)
        if isinstance(parsed, list):
            processed_targets.extend(parsed)
        else:
            processed_targets.append(parsed)
        _LOGGER.debug("Processed Target by json: %s", processed_targets)
    except json.JSONDecodeError:
        if "," in target:
            processed_targets += [item.strip() for item in target.split(",") if item.strip()]
        else:
            processed_targets.append(target.strip())
        _LOGGER.debug("Processed Target by string: %s", processed_targets)
```

Also add a regression test for non-string target pass-through in `tests/test_notify.py`.

## References

- Introduced in: #3446 (by @danielbrunt57)
- Identified in: #3450
- Requested by: @alandtse

SAY-5:
Fixes #3453.

`async_send_message` calls `json.loads(target)` on every entry of the
processed target list without a type guard. `json.loads` only accepts
`str | bytes | bytearray`; any non-string entry (dict, list, int) raises
`TypeError`, which is not caught by the surrounding `except
json.JSONDecodeError` block. The notify call then aborts before reaching
`convert()`.

The fix matches the snippet proposed by @alandtse in the issue:

- Skip non-string targets and append them verbatim (the existing test
  suite already documents this contract under "non-string targets are
  passed through unchanged").
- For string targets, only `.extend()` the parse result when it is a
  list; otherwise append, so a JSON scalar like `'"echo1"'` is not
  iterated into individual characters.

Three regression tests are added to `tests/test_notify.py`:

- `test_non_string_dict_target_does_not_raise`
- `test_non_string_int_target_does_not_raise`
- `test_json_scalar_string_target_appended_not_extended`

`pytest tests/test_notify.py` passes (22 tests).

## Release Notes

* **Bug Fixes**
  * Improved notification service target handling to safely process various target formats, including non-string types and different JSON parsing results.
  * Enhanced robustness in message delivery to prevent type-related errors during target preprocessing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/alandtse/alexa_media_player/pull/3457)

### DIFF
--- a/custom_components/alexa_media/notify.py
+++ b/custom_components/alexa_media/notify.py
@@ -216,8 +216,16 @@ class AlexaNotificationService(BaseNotificationService):
         processed_targets = []
         for target in targets:
             _LOGGER.debug("Processing: %s", target)
+            if not isinstance(target, str):
+                processed_targets.append(target)
+                _LOGGER.debug("Processed non-string target: %s", processed_targets)
+                continue
             try:
-                processed_targets += json.loads(target)
+                parsed = json.loads(target)
+                if isinstance(parsed, list):
+                    processed_targets.extend(parsed)
+                else:
+                    processed_targets.append(parsed)
                 _LOGGER.debug("Processed Target by json: %s", processed_targets)
             except json.JSONDecodeError:
                 if "," in target:
--- a/tests/test_notify.py
+++ b/tests/test_notify.py
@@ -301,6 +301,71 @@ class TestAsyncSendMessageGroupExpansion:
         assert "media_player.echo1" in expanded
         assert "media_player.echo2" in expanded
 
+    @pytest.mark.asyncio
+    async def test_non_string_dict_target_does_not_raise(self):
+        """A dict target must not raise TypeError from json.loads.
+
+        Regression test for issue #3453: ``json.loads`` only accepts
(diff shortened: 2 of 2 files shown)


alandtse (approved):
Thanks!

PR #3469, opened 2026-05-17
fix: preserve cookie files during boot reauth and allow auto-relogin

ksaye:
## Problem

After Home Assistant restarts, the Alexa Media Player integration frequently requires manual re-authentication through the UI (clicking "Next" repeatedly) even though all credentials — email, password, OTP secret, and OAuth tokens — are already stored.

## Root Cause

Two bugs in the boot/reauth flow combine to cause this:

### Bug 1: Cookie files destroyed on boot (`__init__.py`)

When bootstrap cookie auth fails (common after Amazon expires the session), `login.login(cookies=cookies)` is called. Inside alexapy, this path:
1. Calls `test_loggedin(cookies)` → fails
2. Calls `reset()` → **destructively deletes ALL cookie files** via `os.remove()`
3. Falls through to credential login (which also fails without browser proxy)

The cookie pickle file goes from ~5,746 bytes (valid session) to being completely deleted. Even after manual re-auth, the next restart repeats the cycle.

### Bug 2: 60-second reauth guard blocks auto-relogin (`config_flow.py`)

`async_step_reauth` checks if `seconds_since_login < 60` and forces a manual login form when true. On boot, reauth always triggers within ~30 seconds, so the automatic relogin path (`async_step_user_legacy` with stored credentials + TOTP) is **never reached**.

### Fix 1: Don't pass cookies to `login()` when bootstrap already failed

Changed `login.login(cookies=cookies)` → `login.login()`. This skips the `test_loggedin()` → `reset()` path, preserving cookie files for future boots while still attempting OAuth/credential login.

### Fix 2: Always attempt automatic relogin

Removed the 60-second guard that forced manual login. Now both boot-time reauth (< 60s) and runtime reauth (> 60s) flow through `async_step_user_legacy`, which uses stored credentials and the OTP secret for automatic re-authentication.

## Evidence

- **Before restart**: Cookie pickle = 5,746 bytes with valid Amazon session cookies
- **After restart (before fix)**: Cookie pickle = FILE DELETED (not just empty)
- **Boot log (before fix)**: `"Relogin requested within 30 seconds; manual login required"`
- **After fix**: Integration loads all 93 entities without reauth prompt
- **User observation**: "Just clicking next next next" works — proving all needed credentials are already stored

## Testing

- Applied fix to a live Home Assistant instance (HA OS 2026.5.2, Alexa Media Player 5.15.2)
- Restarted HA after applying the fix
- Integration loaded successfully with all 93 Alexa entities — no reauth notification

## Summary by CodeRabbit

* **Bug Fixes**
  * Startup fallback now attempts OAuth token recovery before falling back to form login.
  * Clears cached request state to prevent replay of stale pages during reauth.
  * Boot-time reauth proceeds automatically (no manual prompt) when detected.

* **New Features**
  * Setup now proactively tries to refresh OAuth tokens when available; refresh failures are logged non‑fatally and setup continues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/alandtse/alexa_media_player/pull/3469?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/custom_components/alexa_media/__init__.py
+++ b/custom_components/alexa_media/__init__.py
@@ -673,7 +673,33 @@ async def async_setup_entry(hass, config_entry):
                 hass.data[DATA_ALEXAMEDIA]["accounts"][email]["login_obj"] = login_obj
             else:
                 login_obj.oauth_login = True
-            await login_obj.reset()
+            # Try OAuth token refresh first instead of reset(), which
+            # destructively deletes all cookie files. The HTTP/2 timeout
+            # that triggers relogin is often transient and doesn't mean
+            # the session is invalid.
+            relogin_ok = False
+            try:
+                if login_obj._session is None or getattr(
+                    login_obj._session, "closed", False
+                ):
+                    login_obj._create_session(True)
+                token_ok = await login_obj.get_tokens()
+                if token_ok:
+                    cookie_ok = await login_obj.exchange_token_for_cookies()
+                    if cookie_ok:
+                        await login_obj.get_csrf()
+                        _LOGGER.debug(
+                            "%s: Relogin via OAuth token refresh succeeded",
+                            hide_email(email),
+                        )
+                        relogin_ok = True
+            except Exception:  # noqa: BLE001
+                _LOGGER.debug(
+                    "%s: OAuth refresh failed during relogin, falling back to reset",
+                    hide_email(email),
+                )
+            if not relogin_ok:
+                await login_obj.reset()
             # await login_obj.login()
             if await test_login_status(hass, config_entry, login_obj):
                 await setup_alexa(hass, config_entry, login_obj)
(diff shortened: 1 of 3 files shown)


alandtse:
This is how you get amazon to ban you. Hammering an unofficial API with bad credentials is not a good idea. 

Separately, if you fail to log in, the credentials need to be treated as invalid. If you believe there's a refresh bug, then submit a PR on that. 

Thanks.

ksaye:
> This is how you get amazon to ban you. Hammering an unofficial API with bad credentials is not a good idea.
> 
> Separately, if you fail to log in, the credentials need to be treated as invalid. If you believe there's a refresh bug, then submit a PR on that.
> 
> Thanks.

sorry, was doing some coding with Claude to fix an issue.  I defer to your judgement about the correct steps.  :)

alandtse:
Thanks. I'll close it. I do think we could have a refresh bug that needs to be squashed but per the reasons above, I don't think this is the right fix.

danielbrunt57:
`alexapy` MR !466 created: [fix: Preserve cookie files when cookie login validation fails](https://gitlab.com/keatontaylor/alexapy/-/merge_requests/466)

danielbrunt57:
@ksaye Can you test the PR's [`alexalogin.py`](https://github.com/user-attachments/files/28008699/alexalogin.py)?

> When `login(cookies=...)` fails cookie-based session validation, the login flow
> calls `reset()`, which deletes all persisted cookie files before falling back to
> credential login.
> 
> This can unnecessarily remove otherwise useful persisted session state during
> startup or transient authentication failures.
> 
> `AlexaLogin.login()` currently does:
> 
> ```py
> if cookies:
>     if await self.test_loggedin(cookies):
>         return
>     await self.reset()
> ```
> `reset()` clears in-memory login state and deletes all persisted cookie files.
> 
> As a result, any failed cookie validation destroys the stored cookie jar even
> when the failure may only be transient or recoverable.
> 
> This PR makes `reset()` optionally preserve persisted cookie files and uses the
> non-destructive reset path after failed cookie validation in `login()`.
> 
> This still clears in-memory session/login state before continuing with
> credential login, but preserves persisted cookies unless an explicit destructive
> reset is required.
> 
> #### Notes
> 
> This change does not alter login retry behavior or automatic reauthentication
> policy. It only separates session-state reset from persisted cookie deletion.

PR #3524, opened 2026-07-19
fix: strip whitespace from Amazon region domain

ericchan:
## Summary
- strip whitespace from the Amazon region domain before saving config flow input
- add a regression test for an accidental space in `amazon.com.au`

## Why
A mobile browser/input path can submit `amazon.com.au` as `amazon.com. au`, which then causes alexapy to try `alexa.amazon.com. au` and fail DNS resolution with `Misformatted domain name`.

Removing whitespace from this domain field keeps the existing behavior for valid domains while tolerating accidental spaces from paste/autofill/mobile input.

## Testing
- `python3 -m py_compile custom_components/alexa_media/config_flow.py tests/test_config_flow.py`

I could not run pytest in this local shell because the repo's Python/Home Assistant test dependencies are not installed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved setup handling for Amazon region URLs by cleaning accidental whitespace before saving.
  * Prevents malformed entries (for example, `amazon.com. au`) from being stored, reducing connection and configuration issues caused by copy/paste input mistakes.

### DIFF
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -820,6 +820,7 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
         if CONF_PASSWORD in user_input:
             self.config[CONF_PASSWORD] = user_input[CONF_PASSWORD]
         if CONF_URL in user_input:
+            # Remove accidental whitespace introduced by mobile keyboards/paste.
             self.config[CONF_URL] = "".join(user_input[CONF_URL].split())
         if CONF_PUBLIC_URL in user_input:
             if not user_input[CONF_PUBLIC_URL].endswith("/"):
--- a/custom_components/alexa_media/config_flow.py
+++ b/custom_components/alexa_media/config_flow.py
@@ -820,7 +820,7 @@ class AlexaMediaFlowHandler(config_entries.ConfigFlow):
         if CONF_PASSWORD in user_input:
             self.config[CONF_PASSWORD] = user_input[CONF_PASSWORD]
         if CONF_URL in user_input:
-            self.config[CONF_URL] = user_input[CONF_URL]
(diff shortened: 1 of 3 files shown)


custom_components/alexa_media/config_flow.py
```diff
@@ -820,7 +820,7 @@ def _save_user_input_to_config(self, user_input=None) -> None:
         if CONF_PASSWORD in user_input:
             self.config[CONF_PASSWORD] = user_input[CONF_PASSWORD]
         if CONF_URL in user_input:
-            self.config[CONF_URL] = user_input[CONF_URL]
+            self.config[CONF_URL] = "".join(user_input[CONF_URL].split())
```
danielbrunt57, line 824: Perhaps a short comment would help clarify the intent here, e.g.
```
# Remove accidental whitespace introduced by mobile keyboards/paste.
self.config[CONF_URL] = "".join(user_input[CONF_URL].split())
```

danielbrunt57 (commented):
Looks good overall. Just one small suggestion to make the intent of the whitespace normalization a little clearer.

ericchan:
Added the code comment, thanks.

For context, this came from a real setup failure on Home Assistant where the region field ended up as `amazon.com. au` with an accidental/auto-inserted space, even when entering `amazon.com.au` or `amazon.com`. Alexa Media Player then saved that value as-is and attempted to use it as the Amazon domain, which broke the login flow before it could reach the correct AU endpoint.

This normalization does not change the selected region logically; it just strips accidental whitespace from a domain-style field before saving it. In that case it corrected `amazon.com. au` back to `amazon.com.au`. Valid Amazon domains should not contain whitespace, so this should be low risk while handling bad pasted/autofilled input.

alandtse (approved):

PR #3561, opened 2026-09-11
fix: ignore non-object http2 push messages

### LINKED ISSUE #3560: [Bug]: http2_handler raises AttributeError on push messages that are valid JSON but not objects
## Describe the bug

`http2_handler` assumes every pushed message is a JSON object and calls `.get()` on it
directly. When Amazon's http2 push stream contains a line that is valid JSON but not an
object, the handler raises and the message is dropped:

```
Error doing job: Task exception was never retrieved (task: None)
  File "/config/custom_components/alexa_media/__init__.py", line 2501, in http2_handler
    message_obj.get("directive", {})
    ^^^^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'get'
```

This is not an occasional glitch. On a single instance it fired 518 times in six days,
spread across every day (68 / 138 / 5 / 213 / 82 / 12), accounting for 58 % of all ERROR
lines in the log. Alexa itself keeps working, so the practical impact is lost push updates
plus a log that becomes hard to use for anything else.

## Root cause

The contract between alexapy and the integration is not enforced on either side.

alexapy `alexahttp2.py`, `on_message()` (1.29.25):

```python
elif line and not line.startswith(self.boundary):
    with contextlib.suppress(json.decoder.JSONDecodeError):
        self._schedule(self.msg_callback(json.loads(line)))
```

Every stream line that is not a boundary, a `Content-Type:` header or the reauth notice is
passed through `json.loads()` and handed to the callback unchanged. Only `JSONDecodeError`
is suppressed, so any line that is *valid* JSON reaches the handler regardless of its type.
`json.loads("42")` returns `int`, `json.loads("null")` returns `None`, `json.loads("[1,2]")`
returns `list`, and none of those have `.get()`.

`custom_components/alexa_media/__init__.py` (5.15.7, unchanged in current master):

```python
2382:  async def http2_handler(message_obj):
...
2500:      updates = (
2501:          message_obj.get("directive", {})
2502:          .get("payload", {})
2503:          .get("renderingUpdates", [])
2504:      )
```

No type annotation, no guard. None of the currently open PRs touch these lines.

## Expected behavior

A push message that is not a JSON object should be skipped, optionally logged at debug
level, instead of raising an unretrieved task exception.

## Suggested fix

A guard at the top of the handler:

```python
async def http2_handler(message_obj) -> None:
    if not isinstance(message_obj, dict):
        _LOGGER.debug("Ignoring non-object http2 push message: %r", message_obj)
        return
```

Fixing it in alexapy instead, by dispatching only `dict` results, would protect every
consumer of the library. The guard in the handler is the smaller and more local change.

## To Reproduce

Not reproducible on demand, since it depends on what Amazon pushes. It occurs continuously
on my account with no user interaction. I have not captured the offending raw line yet;
alexapy logs it at debug level in `on_message` ("Received raw message: %s"). Happy to enable
debug logging and attach the payload if that would help pin down which message type it is.

## Versions

- Home Assistant: 2026.9.1
- Alexa Media Player: 5.15.7
- alexapy: 1.29.25
- Amazon domain: amazon.de
- HA core "Alexa Devices" integration also installed: yes
- 52-character key for Amazon 2SV entered: yes

Login and authentication are working normally; this report is not about auth.

dajiaohuang:
## Summary
- ignore HTTP/2 push messages that are not JSON objects before the directive parsing chain in `http2_handler`
- add regression cases for int, float, str, bool, null, list and object messages

## Why
`http2_handler` assumed every alexapy push callback was a JSON object and chained `.get()` on it. alexapy passes the raw `json.loads(line)` result straight to the callback (`msg_callback: Callable[[Any], ...]` in `alexahttp2.py`), so a stream line that is valid JSON but not an object - `123`, `"text"`, `null`, `[1, 2]` - raised `AttributeError` on `.get("directive", {})`, dropping the push update and logging an unretrieved task exception.

Only non-object messages are ignored; object messages take the unchanged path, so genuine failures on valid objects are still surfaced.

Fixes #3560

## Testing
- `pytest tests/` - 212 passed, 2 pre-existing failures in `tests/test_diagnostics.py` (`RuntimeError: Frame helper not set up`), reproduced identically on the unmodified base commit.
- `pytest tests/test_helpers.py -k is_push_message_object` - 12 passed. Mutating the guard to also accept lists fails exactly the list cases.
- `black --check`, `isort --check`, and `pre-commit run` (black, isort, codespell, bandit, pyupgrade, prettier) pass on the changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of unexpected Alexa push message formats.
  - Non-object messages are safely ignored instead of causing processing errors or affecting activity tracking.

- **Tests**
  - Added coverage for valid object messages and unsupported values, including lists, scalars, strings, booleans, and empty values.


custom_components/alexa_media/helpers.py
```diff
@@ -516,6 +516,15 @@ def is_http2_enabled(hass: HomeAssistant | None, login_email: str) -> bool:
     return False
 
 
+def is_push_message_object(message_obj: Any) -> bool:
+    """Whether an http2 push message is a JSON object.
+
+    alexapy hands every stream line that parses as JSON to the callback, so
+    scalars, lists and null reach the handler and have no .get().
+    """
+    return isinstance(message_obj, dict)
```
alandtse, line 525: Does this helper make sense given it's just a isinstance check?

danielbrunt57: @alandtse I agree. It ultimately just wraps isinstance(..., dict). It is only used here and adds unnecessary indirection. The check should be moved directly into http2_handler.  I would keep the regression coverage, but test the handler behaviour rather than the trivial helper and keep the regression coverage focused on the handler ignoring non-object JSON values. It probably only need a small parameterized set:
```
@pytest.mark.parametrize(
    "message_obj",
    [
        None,
        123,
        "text",
        [1, 2],
    ],
)
```

alandtse: Since we handled it in alexapy, I'll close this and if we want the tester, we can add it as a `test` pr.

alandtse:
Thanks. Should be handled by alexapy bump

PR #3553, opened 2026-09-03
fix(switch): drop deprecated via_device key from AlexaMediaSwitch.device_info

simonyang08:
### Problem

Home Assistant 2026.9 emits a deprecation report whenever an integration passes `via_device=` to `device_registry.async_get_or_create`, and HA 2027.8 removes that kwarg entirely. The report surfaces at `custom_components/alexa_media/helpers.py, line 217` (which matches current `dev` — the call site is `add_devices_callback(devices, False)` threading through to the entity platform).

### Why drop the key instead of switching to `via_device_id`

`via_device` accepts a `(domain, identifier)` tuple and asks HA to parent the device under the registry entry with that identifier. `AlexaMediaSwitch.device_info` passed the same `(ALEXA_DOMAIN, client.unique_id)` tuple that already populates `identifiers` — i.e. a self-reference, which HA cannot represent (it is ignored with a log entry today and raises once removed in 2027.8; HA also already prunes self-referencing `via_device_id` values from storage). `via_device_id` is not a drop-in replacement here because the entity has no registry entry id of its own at `device_info` time — that would be circular.

Dropping the key therefore:
* removes the deprecation warning,
* keeps the integration working on HA 2027.8,
* has no observable effect on the device tree (the integration has no hub/bridge concept — every Echo device stands alone, and no other `device_info` in this integration sets `via_device`).

### Test

`tests/test_via_device.py` asserts:

1. `AlexaMediaSwitch.device_info` no longer contains a `"via_device"` key (the actual fix);
2. `identifiers` is still present (sentinel against over-deletion);
3. no file under `custom_components/alexa_media/` re-introduces a `"via_device"` literal (regression sweep).

Two of the three fail on unpatched `dev`; all pass on this branch.

Closes #3551

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed the deprecated self-referential device relationship from Alexa Media switch device information.
  - Device identifiers continue to be included for proper device recognition.

- **Tests**
  - Added regression coverage to verify deprecated device relationship data is not exposed.

Issue #3551, opened 2026-09-02
[WARNING]: Deprecated `via_device` parameter


blue-bean:
### Describe the bug

New warning with core 2026.9.0

`Detected that custom integration 'alexa_media' calls `device_registry.async_get_or_create` with a deprecated `via_device` parameter; use `via_device_id` instead at custom_components/alexa_media/helpers.py, line 217: add_devices_callback(devices, False). This will stop working in Home Assistant 2027.8.0, please create a bug report at https://github.com/alandtse/alexa_media_player/issues`

### To Reproduce

See core log

### Expected behavior

No warning

### Home Assistant Version

2026.9.0

### Alexa Media Player Version

5.5.17

### Alexa Media Player API library Version

?

### Have you entered a 52-character key for Amazon 2SV via Authenticator App login?

No

### Is the HA core integration Alexa Devices installed?

No

### Amazon Domain

amazon.com (US)

### Debug Logs (alexa_media & alexapy)

```text

```

bonkersbrains:
Latest Core update and latest Alexa Media (5.16.0) integration update gets the following error.
Installation method Home Assistant OS
Core 2026.9.2
Supervisor 2026.09.0
Operating System 18.2
Frontend 20260826.7

Logger: homeassistant.components.binary_sensor
Source: helpers/entity_platform.py:725
Integration: Binary sensor ([documentation](https://www.home-assistant.io/integrations/binary_sensor), [issues](https://github.com/home-assistant/core/issues?q=is%3Aissue+is%3Aopen+label%3A%22integration%3A+binary_sensor%22))
First occurred: 11:25:46 AM (2 occurrences)
Last logged: 11:25:46 AM

Error adding entity None for domain binary_sensor with platform alexa_media
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/frame.py", line 190, in report_usage
    integration_frame = get_integration_frame(
        exclude_integrations=exclude_integrations
    )
  File "/usr/src/homeassistant/homeassistant/helpers/frame.py", line 128, in get_integration_frame
    raise MissingIntegrationFrame
homeassistant.helpers.frame.MissingIntegrationFrame

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 725, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
[...]

PR #3531, opened 2026-07-24
feat: add Amazon Kids (child mode) support

chriscomputerfreak:
## Summary
Adds Home Assistant support for **Amazon Kids** (child mode) on Echo devices, using the new alexapy `AlexaAPI` methods (released in alexapy 1.30.0, keatontaylor/alexapy!472).

Closes #349.

## What's added
- **`binary_sensor.<echo>_amazon_kids`** — reflects the per-device `isChildDirectedDevice` state (polled on its own 5-minute interval).
- **`alexa_media.amazon_kids_enable`** (`entity_id`, `child`) — assigns an Echo to a child profile (turns Amazon Kids on). `child` accepts a first name or a directedId.
- **`alexa_media.amazon_kids_disable`** (`entity_id`) — unassigns the device (turns Amazon Kids off).

## Notes
- Bumps `alexapy` to `1.30.0` for the new methods (`get_child_mode` / `disable_child_mode` / `get_child_profiles` / `get_device_child` / `enable_child_mode`).
- The binary_sensor polls per device independently of the main coordinator.

## Testing
- `black` / `isort` clean; rebased on current `dev`.
- Loaded and verified live in a real HA instance: alexapy auto-upgrades to 1.30.0, all 7 Echo sensors report the correct per-device state, and the enable/disable services work end-to-end (incl. the parent-dashboard CSRF bootstrap for enable).

## Summary by CodeRabbit

* **New Features**
  * Added Amazon Kids (child mode) status sensors for supported Echo devices.
  * Added services to enable or disable Amazon Kids on selected Alexa devices.
  * Enable supports choosing a child profile by first name (or directed profile id).
  * Child mode sensor states update periodically and only show availability after the first successful fetch.
* **Documentation**
  * Added user-facing service descriptions and required fields for the Amazon Kids enable/disable actions.
* **Chores**
  * Updated the underlying `alexapy` dependency to a newer version.

Issue #3565, opened 2026-09-12
[Bug]: 'AmazonKidsSensor' object has no attribute 'alexa_entity_id'


Matsuo3rd:
### Describe the bug

Every 10 minutes the following ERROR log appears:

```
2026-09-12 14:39:11.272 ERROR (MainThread) [custom_components.alexa_media.coordinator] Unexpected error fetching alexa_media data
Traceback (most recent call last):
  File "/srv/homeassistant/lib/python3.14/site-packages/homeassistant/helpers/update_coordinator.py", line 441, in _async_refresh
    self.data = await self._async_update_data()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/homeassistant/lib/python3.14/site-packages/homeassistant/helpers/update_coordinator.py", line 315, in _async_update_data
    return await self.update_method()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/homeassistant/.homeassistant/custom_components/alexa_media/__init__.py", line 977, in async_update_data
    entities_to_monitor.add(binary_sensor.alexa_entity_id)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'AmazonKidsSensor' object has no attribute 'alexa_entity_id'
```

### To Reproduce

ERROR log just appears without any specific action

### Expected behavior

Log shall not be showing up

### Home Assistant Version

2026.9.2

### Alexa Media Player Version

5.16.0

### Alexa Media Player API library Version

1.30.0

### Have you entered a 52-character key for Amazon 2SV via Authenticator App login?

No

### Is the HA core integration Alexa Devices installed?

No

### Amazon Domain

amazon.de (Germany)

### Debug Logs (alexa_media & alexapy)

```text

```

BastiPrivat:
Confirmed with Home Assistant OS 18.2, Core 2026.9.2, Alexa Media Player 5.16.0 and alexapy 1.30.0.

`AmazonKidsSensor` instances are appended to the same `entities["binary_sensor"]` collection as Smart Home contact sensors. Kids sensors poll independently and have no `alexa_entity_id`, but the coordinator reads that attribute from every enabled binary sensor. This aborts the entire coordinator refresh. The accompanying unawaited-coroutine warnings follow because the API coroutine list has already been constructed when the exception is raised.

This minimal guard in `async_update_data` restored polling locally while preserving enabled contact sensors:

```python
[...]

PR #3566, opened 2026-09-13
fix(binary_sensor): drop self-referential via_device so Amazon Kids sensors are created

davidcoulson:
### Problem

`AmazonKidsSensor.device_info` declares `via_device` with the **same** `(DATA_ALEXAMEDIA, unique_id)` tuple as `identifiers`, so the device is its own parent:

```python
return {
    "identifiers": {(DATA_ALEXAMEDIA, self._client.unique_id)},
    "via_device": (DATA_ALEXAMEDIA, self._client.unique_id),   # same tuple
}
```

Since HA 2026.9, `device_registry.async_get_or_create` reports the deprecated `via_device` parameter through `frame.report_usage`. For most call sites that is just the warning tracked in #3551 — but the Amazon Kids sensors are added from a task with no integration frame on the stack, so `report_usage` hits `MissingIntegrationFrame` and re-raises it as a `RuntimeError`, which aborts the entity add entirely:

```
ERROR (MainThread) [homeassistant.components.binary_sensor] Error adding entity None
for domain binary_sensor with platform alexa_media
...
  File "homeassistant/helpers/frame.py", line 128, in get_integration_frame
    raise MissingIntegrationFrame
...
  File "homeassistant/helpers/frame.py", line 386, in _report_usage_no_integration
    raise RuntimeError(msg) from err
RuntimeError: Detected code that calls `device_registry.async_get_or_create` with a
deprecated `via_device` parameter; use `via_device_id` instead.
```

So this is not only a deprecation warning — on HA 2026.9+ the Amazon Kids sensors added in #3531 are silently missing.

### Fix

Drop the key. `identifiers` already attaches the sensor to the Echo device created by the `media_player` platform, so the self-referential `via_device` was a no-op; there is no parent to resolve and no `via_device_id` lookup is needed. One deleted line, no behaviour change beyond the entity now being created.

### Testing

HA 2026.9.2, Python 3.14.6, HAOS 18.2, alexa_media 5.16.0, 17 Echo devices, real Amazon account.

| | before | after |
|---|---|---|
| `Error adding entity None ... platform alexa_media` per start | 9 | **0** |
| Amazon Kids binary_sensors present | 1 | **10** |
| total alexa_media entities | 109 | 118 |
| alexa_media `via_device` deprecation warnings | 1 | **0** |

All 10 sensors report a real state (`off`, not `unavailable`), one per Kids-capable Echo. Verified across a full HA restart, not just a config-entry reload, so the module was genuinely re-imported. Switch, sensor and media_player entity counts are unchanged.

### Notes

- Refs #3551.
- `AlexaMediaSwitch.device_info` has the identical self-referential pattern. That half is already covered by the draft in #3553, so I have deliberately left `switch.py` untouched here to avoid duplicating that work — this PR is only the binary_sensor half, which is the one that actually loses entities.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Amazon Kids sensor device information to no longer display an associated Echo device, while retaining the sensor’s device identification.


danielbrunt57:
Superseded by #3570

davidcoulson:
Sounds good — thanks @danielbrunt57. #3570 covers both occurrences, including the `switch.py` one I had deliberately left out of this PR to avoid colliding with the draft in #3553, so it is the better shape. No duplicated effort in flight from my side.

If it is useful for #3570, the verification from this branch carries over unchanged, since the diff is the same: on HA 2026.9.2 with 17 Echo devices, the Amazon Kids sensors went from 1 to 10 and the `Error adding entity None for domain binary_sensor with platform alexa_media` errors from 9 per start to 0, confirmed across a full restart rather than a config entry reload — the module is cached in `sys.modules`, so a reload alone still runs the old code and will look like a false pass.

Separately, #3567 fixes a second bug in the same Amazon Kids feature (#3565): once these sensors actually get created, `AmazonKidsSensor` has no `alexa_entity_id`, which aborts every coordinator refresh. The two land better together — otherwise fixing this one is what exposes that one.

PR #3568, opened 2026-09-13
fix: release hass.data on unload and re-arm metrics

davidcoulson:
### Problem

Two related lifecycle issues.

**1. `hass.data` is never released.** `async_unload_entry` pops `accounts`, `services` and `config_flows`, but never `metrics` or `notify_service`. `DATA_ALEXAMEDIA` therefore stays truthy, the final cleanup branch never runs, and every unload logs:

```
Unable to remove alexa_media data structure: {'metrics': <AlexaMetrics object at 0x...>}
```

The collector and the notify service reference survive for the life of the HA run, across any number of reloads.

**2. Boot metrics are measured from the wrong origin.** `BootMetrics.start_time` is set when the collector is constructed — in `async_setup`, which runs once per HA run — and `record_stage` reports `time.monotonic() - start_time`. Because the collector survives (1), every stage recorded after a reload is measured from the *original* domain setup rather than from the reload.

The values are logged as "completed in", so they read as per-stage durations. Observed on a reload where the same setup's own `[BOOT]` lines reported login 0.58s, http2 0.29s, API fetch 3.16s, `setup_entry total: 5.81s`:

```
[BOOT METRICS] setup_alexa_start_... completed in 175.216s
[BOOT METRICS] platforms_loaded_...  completed in 180.427s
[BOOT METRICS] last_called_...       completed in 180.444s
```

Anyone diagnosing a slow setup from these numbers is off by a factor of 30.

### Fix

- Drop `metrics` and `notify_service` when the last account unloads, so the data structure is actually removed. `notify.async_unload_entry` already removes its HA services at exactly that point but never clears its `hass.data` reference, so this is consistent.
- Re-arm boot tracking at the start of each `async_setup_entry`, recreating the collector if a previous unload removed it.

Every `get_metrics()` call site is already `if metrics:` guarded, so the collector being absent between unload and setup is safe.

### Testing

HA 2026.9.2. Config entry reload, before and after:

| | before | after |
|---|---|---|
| unload message | `Unable to remove alexa_media data structure: {'metrics': ...}` | `Removing alexa_media data structure` |
| `[BOOT METRICS] platforms_loaded` | 180.427s | **8.632s** |
| `[BOOT METRICS] last_called` | 180.444s | **8.639s** |
| actual `[BOOT] setup_entry total` | 5.81s | 8.64s |

After the change the metrics agree with the `[BOOT]` timings for the same setup, `[BOOT METRICS] Started tracking` appears on each reload, and a full restart plus reload cycle produces no alexa_media errors.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved setup tracking reliability for Alexa Media integrations.
  - Preserved monitoring metrics when multiple entries are loaded.
  - Ensured final account removal clears associated metrics and notification-service state.


davidcoulson:
Thanks — the multi-account point was correct, and I've pushed a fix.

Calling `start_boot_tracking()` unconditionally in `async_setup_entry` did replace `AlexaMetrics.boot_metrics` on every entry setup: with two accounts loaded, setting up the second discarded the stages already recorded for the first, and first-entry stages still in flight were recorded against the second entry's tracker.

I took the second of the two options suggested — resetting only at the domain lifecycle boundary — rather than per-entry trackers, because it needs no change to the `metrics.py` API or its call sites. The collector is now removed when the last account unloads, so "the collector is missing" *is* the domain boundary, and re-arming only when it is created gives:

- reload of the only account → collector was popped by the unload → re-armed → measures its own setup (the case this PR is about)
- adding or reloading a second account → collector still present → not re-armed → the loaded entry's stages are left intact

Re-verified on HA 2026.9.2 after the change — a reload still logs `[BOOT METRICS] Started tracking`, and the stage values agree with the `[BOOT]` lines for the same setup (`last_called ... 7.07s` against `[BOOT] setup_entry total: 7.08s`).

alandtse (approved):

PR #3569, opened 2026-09-13
fix: bound AlexaLogin.login() timeout

davidcoulson:
### Problem

`async_setup_entry` awaits `login.login()` with no overall bound:

```python
if not cookie_login_ok:
    await login.login(cookies=cookies)
```

`AlexaLogin.login()` is a long multi-step state machine (~1200 lines of sequential requests in alexapy 1.30.0), and the `aiohttp.ClientSession` it uses is constructed without a `timeout=`, so individual requests fall back to aiohttp's default while nothing limits the call as a whole.

If it stalls, `async_setup_entry` never returns. Home Assistant has no maximum wait for config entry setup, so the entry sits in "setting up" indefinitely — no `ConfigEntryNotReady`, no retry, nothing surfaced in the UI, and nothing logged above debug. The only visible symptom is that the integration's entities never appear.

### Fix

Wrap the call in the `async_timeout` idiom already used elsewhere in this module (`async with async_timeout.timeout(45)`) and translate a timeout into `ConfigEntryNotReady`, so HA applies its normal retry-with-backoff and shows the entry as retrying.

`LOGIN_MAX_WAIT_S` is deliberately 600s: this is a **deadlock guard, not a latency budget**. It is far longer than any healthy login, so it should never fire for a working account, and a slow-but-successful login is unaffected.

### Honesty about scope

I have **not** reproduced a stalled login. This guards a path that currently has no way to fail safely, rather than fixing an observed hang — I went looking after chasing a "slow to start" report that turned out to have a different cause entirely. If you would rather not add a timeout here, or want a different bound, I am happy to adjust or close this.

The other two PRs I opened today (#3567, #3568) are for reproduced, measured bugs; this one is the speculative one of the set.

### Testing

HA 2026.9.2: full restart plus config entry reload with the change in place — startup completes in 30.48s, login path unaffected (`login completed in 0.68s` via the cached-cookie fast path), no alexa_media errors. The timeout branch itself is not exercised by a healthy login, by design.

## Summary by CodeRabbit

* **Bug Fixes**
  * Account login during setup now has a maximum wait time, preventing setup from hanging indefinitely.
  * When login times out, setup stops with a clear timeout message and cleans up the interrupted login session.
  * Retrying setup now starts with a fresh login session.


davidcoulson:
Thanks — the interrupted-client point was right, and it mattered more than the timeout itself. Pushed a fix.

Confirmed against `alexapy==1.30.0`: `AlexaLogin.login()` resumes from `self._lastreq` when one is present (`alexalogin.py` ~L1137-1141), so a login interrupted by the timeout can leave the client continuing a dead request. And `async_setup_entry` hands the retry that same client from **two** places — the stored `hass.data[...]["login_obj"]` and `config_entry.runtime_data` — neither of which is cleared when setup raises `ConfigEntryNotReady`. As written, my guard would have turned a transient stall into a persistent failure, which is worse than the unbounded wait it was meant to protect against.

The timeout handler now closes the old client, pops `login_obj`, and clears `runtime_data` so the retry constructs a fresh `AlexaLogin`.

One deliberate choice worth flagging for review: I did **not** use `AlexaLogin.reset()`, even though it is the obvious cleanup call and is already used elsewhere in this module. `reset()` ends with `delete_cookie()` over `self._cookiefile`, so it would turn a transient stall into a full re-authentication. Only the in-memory client is discarded; the saved cookie is left in place so the retry can still take the `/api/bootstrap` fast path.

Also trimmed the rationale comment on `LOGIN_MAX_WAIT_S` as suggested.

To repeat what is in the PR description, since it matters for how you weigh this one: **I have not reproduced a stalled login.** This hardens a path that currently has no way to fail safely, found while chasing a "slow to start" report that turned out to have an unrelated cause. If you would rather not bound this call, or want a different value than 600s, I am happy to adjust or close it — no attachment to it landing.

alandtse (approved):

